PR #1, opened 2020-01-27
Business-Site-Project

emlo2018:
Hi,

I tried to replicate the layout in the example provided in canvas. If I had more time I would try to insert the video in the header in different formats to fit all devices. It would be nice to make the form straighter, now its jumps around. Especially the input-fields and labels..

/Emelie


code/index.html
```diff
@@ -1,30 +1,94 @@
...
+      </form>
+    </section>
 
     <section class="accordion">
-      <!-- Accordion -->
+      <!-- Accordion / Toggle-->
       <h1>FAQ</h1>
 
-      <div class="question" id="section1">Section 1</div>
+      <div class="question" id="section1">How much does it cost? </div>
```
ericaglimsholt: Maybe `question-one` would be a nice class name here?

code/script.js
```diff
@@ -7,3 +7,14 @@ function toggle() {
...
+document.getElementById("section2").onclick = toggle
+
+
+/*This will eventually be the toggle for accordion*/
+
+const btn = document.getElementById("button")
+
+btn.onclick = () => {
+  btn.classList.toggle("btn-dark")
+}
```
ericaglimsholt: What is this doing? What I can see it does only makes the button color. You can do it with css only actually! [Here](https://www.w3schools.com/cssref/sel_visited.asp) you read more about it :)

code/style.css
```diff
@@ -1,7 +1,183 @@
+
```
ericaglimsholt: In this file some wierd indention is going on. In some places it's two spaces and in some four spaces. It would be great if the file follows one of them!

code/style.css
```diff
@@ -1,7 +1,183 @@
...
+.accordion {
+  background: rgb(245, 244, 242);
+  color:rgb(105, 105, 151);
+  border-radius: 2px;
+  margin: 15px;
+}
+
+/* active, .question:hover {
+  background-color: rgba(82, 100, 151, 0.404); 
+} */
```
ericaglimsholt: Code that you are not using can be deleted when you are submitting your code as a pull request (as you have done here). Once you submit it as a pull request it should be the final code. I guess you are commenting this out to have it later on, but it you do a commit with this you can check your commits to find the code. So it won't be lost :)

code/style.css
```diff
@@ -1,7 +1,183 @@
...
 this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
 
 .active + .answer {
   /* some styling here */
+  display: block;
+  margin: 0;
+  padding: 20px;
+}
+
+@media (min-width: 650px) and (max-width: 1024px) {
```
ericaglimsholt: The code that inside of `@media`s opening and closing tag should be indented one step, so it's easy to see what does belong to it

PR #3, opened 2020-01-27
Week 3 - Bea

bealun:
If I had more time and energy this week I would have spent it exploring more of the JavaScript elements and maybe create some type of form validtors. Trying to get the transition to work smoothly took a lot of time but was definitly worth the work which I'm super proud of!


code/index.html
```diff
@@ -1,30 +1,92 @@
...
+<html lang="en">
   <head>
     <meta charset="utf-8" />
+    <link rel="stylesheet" href="style.css">
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Project 3</title>
-    <!-- dont forget to add a css file and link it here! -->
+    <link rel="shortcut icon" href="img/cupicon.jpg" type="image/x-icon">
+    <link href="https://fonts.googleapis.com/css?family=Bowlby+One+SC&display=swap" rel="stylesheet">
+    <!-- <link href="https://fonts.googleapis.com/css?family=Shadows+Into+Light&display=swap" rel="stylesheet"> -->
```
ericaglimsholt: We don't need code we don't use 😀
```suggestion
```

code/index.html
```diff
@@ -1,30 +1,92 @@
...
+    </header>
+
 
     <!-- Signup form -->
+  <main class="main">
+    <div class="form">
+      <form action="https://httpbin.org/anything" method="POST">
+
+        <label class="textArea">Your Name:<br>
+        <input type="text" required name="fullname" placeholder="Full Name"><br></label>
```
ericaglimsholt: The `<br>` element is from start a breakpoint being used in text purpose, not for styling purpose. This styling could be fixed with only css, so maybe next time try out creating this styling with only css!

code/index.html
```diff
@@ -1,30 +1,92 @@
...
+
+    </section> 
+  </main>
+
+  <script src="script.js"></script>
+
+</body>
+
+  <footer>
+    <a href="tel: 0171-123 45"><img src="img/phone.jpg" alt="phone"></a>
```
ericaglimsholt: In this case the image doesn't fulfil a purpose for someone that can't see it, it's only here to make the phone link to be pretty. So then you can leave `alt` empty! Then a screen reader would skip to read the image and instead only read the link
```suggestion
    <a href="tel: 0171-123 45"><img src="img/phone.jpg" alt=""></a>
```

code/script.js
```diff
@@ -5,5 +5,27 @@ function toggle() {
...
+// const question2 = document.getElementById("section2")
+// const question3 = document.getElementById("section3")
+
+
+
+{/* <p id="demo" style="display:none">Hello JavaScript!</p>
+
+<button type="button" onclick="document.getElementById('demo').style.display='block'">Click Me!</button> */}
+
+// onclick="toggle('answer')"
```
ericaglimsholt: We can delete code that we don't use!

PR #4, opened 2020-01-27
Project 2 - Johanna Åstrand

jastrand:


style.css
```diff
@@ -0,0 +1,217 @@
+/* the + sign is used to select the sibling that comes immediately next to the element you selected.
+in this case it is the div with the answer class next to the div with the active class.
+this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
+
```
ericaglimsholt: This file doesn't follow the same indention everywhere. It would be great if it's the same everywhere. It will make the code more beautiful and easy to read!

PR #5, opened 2020-01-28
Sara Lovén

Aqveduktis:
website: https://gallant-babbage-3591ed.netlify.com/
fun project :)


index.html
```diff
@@ -0,0 +1,127 @@
...
+          <p>Please fill in this form for more information regarding the class and how to participate.</p>
+          <hr>
+          <label for="name">Name</label>
+          <input type="text" placeholder="First Name Last Name" name="name" required>
+
+          <label for="email">Email</label>
+          <input type="text" placeholder="Enter Email" name="email" required>
+
+          <label for="phone">Phone Number</label>
+          <input type="text" placeholder="+46 10 101 0101" name="phone" required>
```
ericaglimsholt: There is actually a input type called `number` that you can use here!

index.html
```diff
@@ -0,0 +1,127 @@
...
+          <label for="name">Course Level</label>
+          <select name="course" id="course">
+            <option value="b">Beginer</option>
```
ericaglimsholt: A good practice is to use shortcut only when you understand what it means. "b" is kind of hard to understand by its own what it is
```suggestion
            <option value="beginner">Beginner</option>
```

index.html
```diff
@@ -0,0 +1,127 @@
...
+
+          <button type="submit" class="sub-btn">Register</button>
+        </div>
+
+      </form>
+
+    </section>
+    <!-- *****Footer***** -->
+    <footer>
+      <!-- <h1>footer</h1> -->
```
ericaglimsholt: Code we don't use can we delete! 😀
```suggestion
```

style.css
```diff
@@ -0,0 +1,331 @@
+/* the + sign is used to select the sibling that comes immediately next to the element you selected.
+in this case it is the div with the answer class next to the div with the active class.
+this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
+/* 
+.active + .answer {
+  
+} */
```
ericaglimsholt: Code we don't need we can delete!

style.css
```diff
@@ -0,0 +1,331 @@
+/* the + sign is used to select the sibling that comes immediately next to the element you selected.
```
ericaglimsholt: In this file there a some empty rows and the indentions isn't the same everywhere. It would be great if the whole file would behave the same!

style.css
```diff
@@ -0,0 +1,331 @@
...
+}
+.input {
+  width: 50%;
+}
+}
+@media all and (min-width: 992px){
+
+ 
+  
+}
```
ericaglimsholt: If not in use - we can delete this media!

PR #6, opened 2020-01-28
Project Business Site - Lisa Sturegard

LisaStur:
Finished the project, struggled a lot with the video header and getting the text in the right place on top before i realised I should create a "header-content"-class that I could manouver.
Also tried to correct the "name"-command for posting the "Name Lastname" in the form, but could not push (maybe because I had edited the Read Me in GitHub?). Will try again and see if I can get it to work...


code/index.html
```diff
@@ -3,28 +3,84 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Project 3</title>
-    <!-- dont forget to add a css file and link it here! -->
+    <link rel="stylesheet" href="./style.css">
+    <link href="https://fonts.googleapis.com/css?family=Dancing+Script&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css?family=Questrial&display=swap" rel="stylesheet">
```
ericaglimsholt: Tip! You can select multiple fonts in Google fonts, which means that you will only get one `<link>`instead of two!

LisaStur: Big thanks! Very handy! :)

PR #13, opened 2020-01-29
Business site project - Anna Takman

annatakman:
I even managed to incorporate some stretch goals in the project, like the slide effect and the video header. The part that I found most difficult was the positioning of the video in the header, and I am not sure that I understand it completely.


code/style.css
```diff
@@ -1,7 +1,205 @@
-/* the + sign is used to select the sibling that comes immediately next to the element you selected.
```
ericaglimsholt: This file needs some work on the indention! 😀

annatakman: Thank you! What is the CSS indentation supposed to look like?

ericaglimsholt: Same way as the html and js. It should be nested of each other
`.my-class {` is where we declare what element we want to change
`  background-color: red;` inside we do the changes for that element
`}` then we close the element

So it becomes 
```
.my-class {
  background-color: red;
}
```

PR #14, opened 2020-01-29
project-business-site - Linnéa

linneanylen:
I had to make a new pull request. 
(Had some problem with github that my latest changes didn't update.) 

But here we go again:
This week mostly I had struggle with finding the time.
So I had to satisfied with only manage the minimum requirements.
Mostly i struggle with:

Adding media queries to video tag in HTML.
It didn't work in newest chrome version. Therefore I changed to css- styling.
Later on I also changed to a picture and tried out the pictures tag instead.
Another struggle was js, therefore I only succeed to do a simple solution.

If I had more time:

Practice js
Adding conditional coloring on Q/A
Structure the code better
Next week I will have better time and hope that I will have time to practice and try out more.


code/style.css
```diff
@@ -5,3 +5,212 @@ this can be used for the FAQ-accordion. read more here: https://www.w3schools.co
 .active + .answer {
   /* some styling here */
 }
+
```
ericaglimsholt: Hmm... There are some wierd indention going on in this file. It needs some love 😀

PR #16, opened 2020-01-29
Week 3 project

jihyeona:
Hi,

I'm still struggling with these problems;
- making the header background image responsive in mobile version: the width of the image is bit narrower than the screen and I don't know how to fix it.
- transition for the accordion: the code is there in CSS, but it seems not working.

Cheers,
Jihyun


code/index.html
```diff
@@ -1,30 +1,143 @@
...
+          </div>
+        </p>
+
+        <button id="signupbtn" class="btn" type="submit">Register</button>
+      </form>
+    </div>
+
+    <!-- //****** FAQ Accordion ******// -->
+    <div class="container">
+      <h1>Frequently Asked Questions</h1>
```
ericaglimsholt: Try to use only one `<h1>` if possible! A `<h2>` would suit great here

code/index.html
```diff
@@ -1,30 +1,143 @@
...
+  </div>
+  </div>
+  </div>
+
+  </div>
+
+
+  </div>
+
+  <script src="script.js"></script>
```
ericaglimsholt: I think this empty lines and indentions can be worked on a bit!

code/script.js
```diff
@@ -1,9 +1,60 @@
...
+      console.log("window.pageYOffset >= sticky");
+    } else {
+      console.log("Not window.pageYOffset >= sticky");
+    }
+    if (window.pageYOffset >= sticky) {
+      navbar.classList.add('sticky');
+    } else {
+      navbar.classList.remove('sticky');
+    }
+  }
```
ericaglimsholt: Ohh nice that you have done more in javascript! A good way to understand it more. But for making the header fixed, you actually don't need javascript to make it work - it works with only css!

jihyeona:
When I look the page on mobile, the navbar and the background image is not filing the whole screen but maybe 80-90 percent of the width of the screen. The text for the header is also out of proportion.  I don’t know what is causing these 🤔

ericaglimsholt:
> When I look at the page on mobile, the navbar and the background image is not filing the whole screen but maybe 80-90 percent of the width of the screen. The text for the header is also out of proportion. I don’t know what is causing these 🤔

I think you need to take a look at box sizing! :) https://www.w3schools.com/css/css3_box-sizing.asp

PR #22, opened 2020-01-29
Project-business-site

Jenniferot:
Here comes my business site! 

I've kept it simple, no advanced stretch goals since I'm still struggling with the basics.
It took some time to understand the javascript, and I've made it very similar to the codepen assignment. And maybe not as it was prepared in the files we first got. Is that ok?

I tend to build up my css with many layers. Maybe it is possible to simplify it a bit. 

Best from Jennifer :-)


code/index.html
```diff
@@ -1,30 +1,108 @@
...
-    <!-- dont forget to add a css file and link it here! -->
-  </head>
-  <body>
-    <!-- video or image as a header is cool :) -->
 
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Project 3</title>
+  <!-- dont forget to add a css file and link it here! -->
```
ericaglimsholt: You have done it so we can remove the helper comment! 😀
```suggestion
```

code/index.html
```diff
@@ -1,30 +1,108 @@
...
+
+
+    <video controls autoplay>
+      <source src="videos/SampleVideo_360x240_1mb.mp4" media="(max-width: 600px)" type="video/mp4">
+      <source src="videos/SampleVideo_720x480_1mb.mp4" media="(max-width: 900px)" type="video/mp4">
+      <source src="videos/SampleVideo_1280x720_1mb.mp4" type="video/mp4">
+      <p>This browser doesn't support video...</p>
+    </video>
+  -->
+
```
ericaglimsholt: Code that we don't use can be removed before making a pull request. If you want to have it saved you can add it to a commit before removing it. Then it will be easy to access later on!

code/script.js
```diff
@@ -1,9 +1,20 @@
 // A function that adds and remove the class "active" on the section you click on.
 // We haven't really talked about ´this´ yet, but we will... ;)
 // console.log(this) to see how it works
+
```
ericaglimsholt: Same here about the out commented code 😀

PR #23, opened 2020-01-29
Project 3 - Anna G

annaelsamaria:
I think I got the hang of flexbox and how to figure out how to work with the parents and children to make everything the way I wanted. I did some of the stretchgoals but I never had the time to try out video. 

The JavaScript was hard, I hope to understand more in the following weeks.

I have one tiny thing that I cant figure out and that is that my FAQ-section is like 1-2 pixels higher up than the form-section. I've looked in the inspector-tool but I cant figure out why. 

Here is my site:
https://stupefied-wing-ba0358.netlify.com/


code/index.html
```diff
@@ -1,30 +1,97 @@
...
+        I
+        love I love chocolate. Gummi bears I love powder bonbon.</p>
+      <button type="button">INVITE ME</button>
+    </div>
+  </section>
+
+  <!-- ORDER PARTY FORM -->
+  <section class="main-content">
+    <div class="order-form">
+      <h1>MAKE YOUR ORDER</h1>
```
ericaglimsholt: Try to use only one `<h1>` per page if possible!

code/style.css
```diff
@@ -2,6 +2,184 @@
 in this case it is the div with the answer class next to the div with the active class.
 this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
 
+* {
```
ericaglimsholt: I think this file needs some love to it. Not the functionality in the CSS, more structure wise. There is some un needed empty rows and the indention isn't the same everywhere. Those things can seems to be nothing and not important, but it really is! Good looking code is as important as the functionality. And it is something that developer recruiter will look at so it's good to work on it from the start 😀

PR #24, opened 2020-01-29
Project 3: Pål Hamel

palhamel:
I think I've finished all requirements and a few of the stretched goals. As last time, I struggle with stopping at "good enough" but I'm prepared of that now. :)


code/index.html
```diff
@@ -1,30 +1,98 @@
...
+    <div class="box-container">
 
-    <section class="accordion">
-      <!-- Accordion -->
-      <h1>FAQ</h1>
+          <!-- MY MAGIC SIGN-UP FORM -->
+      <section class="my-form">
+        <div>
+          <form action="http://httpbin.org/anything" target="_blank" method="POST">
+            <h2>LET'S STAY IN TOUCH - IT'S ALL ABOUT TREES:</h2>
```
ericaglimsholt: A tip is to write this in normal letters and instead in css make it uppercase. Then it will look better in other pages, like in Google, but it will still look like you want styling wise!

palhamel: Thank you Erica! You are so right about the letters, I missed on that. 🌼

PR #26, opened 2020-01-29
Project 3 - Business site (Johanna)

olsenjohanna:
I did not manage to make the styling for some elements work in Safari on my iPhone (for example for select the styling did not apply in Safari on my iPhone as it did in Chrome on my computer).

I struggled a bit but managed to do some stretch goals. I got some help from Jenny and had some aha moments. The javascript was tricky to get at first but now I understand the basic function we used in this project. Looking forward to learn more!

Link: https://loving-tesla-63200c.netlify.com

### DIFF
--- a/code/script.js
+++ b/code/script.js
@@ -1,9 +1,31 @@
 // A function that adds and remove the class "active" on the section you click on.
 // We haven't really talked about ´this´ yet, but we will... ;)
 // console.log(this) to see how it works
-function toggle() {
+
+
+// function toggleFAQ will toggle between class active and without class active
+// active class is styled in CSS (does not exist in HTML)
+// in CSS .active {} will style the question
+// .active + .answer {} display and style the answer 
+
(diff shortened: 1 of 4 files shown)


code/script.js
```diff
@@ -1,9 +1,31 @@
...
+function toggleFAQ() {
   this.classList.toggle("active")
 }
 
-// Selects an HTML element, and calls a function which will be executed when the element is clicked.
-document.getElementById("section1").onclick = toggle
+
+// variables
+
+var section1 = document.getElementById("section1")
```
ericaglimsholt: If possible try to use 1) `const`, 2) `let` 3) `var` when declaring your variables. For this variable it would be great as a `const`! If you want to know more in detail why, then here's a [article](https://tylermcginnis.com/var-let-const/)!  
```suggestion
const section1 = document.getElementById("section1")
```

PR #30, opened 2020-01-30
Project 3, Ingrid H

ingrid2019-ism:


code/index.html
```diff
@@ -1,30 +1,83 @@
...
-    <!-- video or image as a header is cool :) -->
 
+<head>
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Project 3</title>
+  <!-- dont forget to add a css file and link it here! -->
+  <link rel="stylesheet" href="style.css">
+  <link href="https://fonts.googleapis.com/css?family=Quicksand&display=swap" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/css?family=Dancing+Script&display=swap" rel="stylesheet">
```
ericaglimsholt: Tip! In Google Fonts you can select multiple fonts and then it will give you only one `<link>` instead of two!

code/index.html
```diff
@@ -1,30 +1,83 @@
...
+  <!-- video or image as a header is cool :) -->
+  <header>
+    <h1>Nordic Dance Art</h1>
+  </header>
+
+  <!--Poem beneath the header-->
+  <!--<div>
+    <h2>Life isn´t about waiting for the storm to pass...</h2>
+    <h2>It´s about learning to Dance in the rain</h2>
+  </div>-->
```
ericaglimsholt: Code that we don't use can be deleted before the pull request to make it "clean". If you want to save the code for later you can make a commit with it and after delete it. Then you can go back to that commit later on to get the code!

code/style.css
```diff
@@ -2,6 +2,226 @@
 in this case it is the div with the answer class next to the div with the active class.
 this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
 
+
```
ericaglimsholt: This file could need some more love! There's some empty rows that makes it harder to read (but sometime it helps!) and the indention is not always the same everywhere

PR #31, opened 2020-01-30
Petra Jud "business" site

peju3061:
I am still unsure if I use unnessecary classes and IDs, in cases where I may target the elements by their semantic attributes. I am also unsure whether I tab correctly.


code/index.html
```diff
@@ -3,28 +3,92 @@
...
+
+    <label>WILL YOU JOIN FOR DINNER?
+    <div class="form-answer">
+    <div class="food-answer">
+      <input type="radio" name="dinner" value="yes">YES
+    </div>
+    
+    <div class="food-answer">
+      <input type="radio" name="dinner" value="no">NO
+    </div>
```
ericaglimsholt: This would need one more indention!

PR #32, opened 2020-01-30
Project 3 Bussiness site

Sickan-sys:


code/index.html
```diff
@@ -1,30 +1,93 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="style.css">
+    <title>🐶 Dogpool</title>
+</head>
+<body>
+    <header> 
+      
+        <div class="header-video" style="width: 1349px; height: 800px;">
+             <video autoplay="true" loop="true" muted="" id="header-video"> type="video/mp4">
+            <source src="C:\Users\siri_\Desktop\KodÖvEx\Pexels Videos 2776522.mp4" type="video/mp4"></video></div>
```
ericaglimsholt: The source of the video isn't correct. So you need the path to be from this page you are on. So the video and `index.html` is in the same file so you only need `video.mp4` as your `src`. A tip is to rename your video file to something easier and without spaces!

code/index.html
```diff
@@ -1,30 +1,93 @@
...
+               <option value="Rehab30">Rehab 30 minutes</option>
+               <option value="Rehab60">Rehab 60 minutes</option>
+               <option value="Exercise30">Exercise 30 minutes</option>
+               <option value="Exercise60">Exercise 60 minutes</option>
+             </select>
+             <label>
 
-    <!-- Signup form -->
+             <label>
+               <input type="radio" name="Dogsize" value="BigDog" > Big dog
```
ericaglimsholt:
```suggestion
               <input type="radio" name="Dogsize" value="BigDog"> Big dog
```

code/index.html
```diff
@@ -1,30 +1,93 @@
...
+             <label>
+                <button class="bone">
+                    <div class="c1"></div>
+                    <div class="c2"></div>
+                    <div class="c3"></div>
+                    <div class="c4"></div>
+                    <div class="b1">
+                      <div class="b2">
+                        Submit
+                      </div>
```
ericaglimsholt: Try to name your classes so its more as a description of the element. It's hard to know what `c3` stands for!

code/script.js
```diff
@@ -1,9 +1,16 @@
...
+var i;
+
+for (i = 0; i < acc.length; i++) {
+  acc[i].addEventListener("click", function() {
+    this.classList.toggle("active");
 
-// Selects an HTML element, and calls a function which will be executed when the element is clicked.
-document.getElementById("section1").onclick = toggle
+   
+    var panel = this.nextElementSibling;
```
ericaglimsholt: Try use `const` or `let` before `var` where it's possible!
```suggestion
    const panel = this.nextElementSibling;
```

code/style.css
```diff
@@ -1,7 +1,188 @@
-/* the + sign is used to select the sibling that comes immediately next to the element you selected.
-in this case it is the div with the answer class next to the div with the active class.
-this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
+* {
```
ericaglimsholt: This file needs some work on it's structure! There are some empty rows that can be deleted and it would be awesome if the indention is the same everywhere. Then it will make the code more beautiful and more easy to read, which we always should aim for!

ericaglimsholt:
Hi Siri! Can you update the Netlify link with your new code? :)

PR #33, opened 2020-01-30
Project 3 - Nathalie Rautell

rautellin:
Finished the project and did following stretch goals:

[x] Add an icon before the title in the accordion indicates whether the section is expanded or collapsed. It should update when the section expands or collapses.

[x] Try to add some animations on your project. For example a slide-down effect when the accordion expands and collapses.

[x ] Add a focus effect on the input tags (Using the CSS `:focus` pseudo-selector). Make the border or the background of the field change colour and add a glow effect.

[x] Add some validation to the form.

[x]  Add a header and/or a button on top of your video, just like Confetti, by using position: absolute and z-index to get everything in the right place.

This project was more fun since now I understand css better, however the js code could be cleaner.


code/final/index.html
```diff
@@ -0,0 +1,150 @@
...
+  <link rel="icon" href="icon.png">
+  <title>Hire a developer!</title>
+</head>
+
+<body>
+
+  <header class="v-header header-container">
+    <div class="fullscreen-video-wrap">
+      <video id="myvideo" preload autoplay muted loop playsinline>
+        <!-- source tag goes here with javascript -->
```
ericaglimsholt: You can delete this comment! It's just a helper text for when starting the project ☺️
```suggestion
```

code/final/index.html
```diff
@@ -0,0 +1,150 @@
...
+    </section>
+
+    <div class="grid-container">
+      <form action="http://httpbin.org/anything" method="POST">
+
+        <h4>Contact me</h4>
+        <p>Interested in my services or need advice? Then please get in touch and I'll be glad to help.</p>
+
+        <h4>About you</h4>
+        <!-- the for and id need to have the same value to be connected to each other -->
```
ericaglimsholt: This is also just a help comment!
```suggestion
```

code/final/script.js
```diff
@@ -0,0 +1,36 @@
+var accordions = document.getElementsByClassName("accordion");
```
ericaglimsholt: Try use `const` or `let` before using `var` in those places it works!

PR #35, opened 2020-01-31
project-business-site complete!

elsareutersward:
Hi, Here is my finished project!

My Live page can be found at: https://serene-dubinsky-506c1f.netlify.com

I struggled with the hero-video for a bit, but think I solved it in the end. I didn't think this particular video suited when scaling down to tablet/mobile, why I changed it to a picture for those sizes. 
I got the slide function to work fine on the first accordion (Our dogs) but its not so smooth on the second accordion (FAQ) where it kind of jerks/is choopy.


scripts/script.js
```diff
@@ -0,0 +1,19 @@
+
+// Funktion för att stänga andra
+
+function toggle() {
+  var activeSections = document.getElementsByClassName("active");
```
ericaglimsholt: If possible - try use `const` or `let` before using `var`!
```suggestion
  const activeSections = document.getElementsByClassName("active");
```

scripts/script.js
```diff
@@ -0,0 +1,19 @@
...
+  for (var j = 0; j < activeSections.length; j++) {
```
ericaglimsholt: What does `j` stand for in this context? :)

styles/style.css
```diff
@@ -0,0 +1,527 @@
+/******* Body *******/
+body {
+  margin: 0;
+  background-color: #ff9860;
+  background-image: linear-gradient(0deg, #ff9860 0%, #f3d692 38%, #fdefcd 100%);
+  font-family: 'Roboto Mono', monospace;
+}
+
+/******* Header *******/
```
ericaglimsholt: I like these sections!

styles/style.css
```diff
@@ -0,0 +1,527 @@
...
+  display: grid;
+  grid-template-columns: 10% 36% 8% 36% 10%;
+  grid-template-rows: auto;
+  grid-template-areas: 
+    ". item-a item-a item-a ."
+    ". item-b . item-c .";
+  justify-items: center;
+}
+
+/*______ Accordion Dogs ______*/
```
ericaglimsholt: Ohh here's another way. What's is the different in your file between a section that has comment `/******* Header *******/` and `/*______ Accordion Dogs ______*/`?

styles/style.css
```diff
@@ -0,0 +1,527 @@
...
+  display: block;
+}
+
+.radio-container .radiomark:after {
+ 	top: 7px;
+	left: 7px;
+	width: 6px;
+	height: 6px;
+	border-radius: 50%;
+	background: white;
```
ericaglimsholt: Unindent one step!

PR #39, opened 2020-01-31
project-business-site // Anne-Sophie

annesophiegendron:


code/index.html
```diff
@@ -1,30 +1,129 @@
...
+              <input type="text" id="lastname" name="user_lastname" class="form-control"
+                placeholder="Enter your last name" required>
+            </div>
+            <div class="labels">
+              <label for="mail">E-mail address:</label>
+              <input type="email" id="mail" name="user_mail" class="form-control" placeholder="Enter your email"
+                required>
+            </div><br>
+            <div>
+              <p>Pick a time slot:</p>
```
ericaglimsholt: Maybe an [input with date type ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date)would be suited here?

code/index.html
```diff
@@ -1,30 +1,129 @@
...
+              <input type="password" id="pass" name="password" minlength="8" required>
+            </div><br>
+            <div>
+              <label>
+                <input name="conditions" value="validation" type="checkbox" class="input-checkbox" required />I
+                agree
+                to
+                the
+                Terms
+                and Conditions</label>
```
ericaglimsholt: How every word is on different rows? 😀

code/script.js
```diff
@@ -1,9 +1,30 @@
...
+
+const toggleActiveClassFourth = () => {
+  document.getElementById('q4').classList.toggle('active')
+  document.getElementById('a4').classList.toggle('active')
+}
+
+const toggleActiveClassFifth = () => {
+  document.getElementById('q5').classList.toggle('active')
+  document.getElementById('a5').classList.toggle('active')
+}
```
ericaglimsholt: You could actually narrow the functions down to one that can take all q1-q5. So if you want a general function it can be called for example
```
const toggle = () => {
}
```
Inside your function now you have `document.getElementById('q5').classList.toggle('active')`. We can make that with out general function as well. But we need to pass on `q5` so we know what element we should toggle `active` on.
```
const toggle = (questionId) => {
  console.log(questionId); //output: 5
}

document.getElementById('q5').onclick = toggle(5);
```
We can then use our questionId in your code
```
const toggle = (questionId) => {
  document.getElementById('q' + questionId).classList.toggle('active')
}

document.getElementById('q5').onclick = toggle(5);
```
That's all you need! What you can do next is to add the `a5` as well :)

code/style.css
```diff
@@ -1,7 +1,332 @@
...
+}
+
+h3 {
+  color: black;
+  text-align: center;
+  font-size: 20px;
+}
+
+h4 {
+
```
ericaglimsholt:
```suggestion
```

code/style.css
```diff
@@ -1,7 +1,332 @@
...
+font-size: 15px;
+font-weight: lighter;
```
ericaglimsholt: Indent

code/style.css
```diff
@@ -1,7 +1,332 @@
...
+
+.head-video video {
+  position: relative;
+  z-index: 0;
+  width: 100%;
+  height: 100%;
+  flex-grow: 1;
+  opacity: 0.98;
+  margin-top: -3px;
+  }
```
ericaglimsholt:
```suggestion
}
```

code/style.css
```diff
@@ -1,7 +1,332 @@
...
+  transition: all 0.5s ease;
+}
+
+.answer.active {
+  height: 30px;
+  transition: all 0.5s ease;
+  overflow: scroll;
+}
+
+/****************** FOOTER ******************/
```
ericaglimsholt: I like these sections!

PR #43, opened 2020-09-02
Project Business Site - Vanessa Sue Smith

VanessaSue27:
I decided to use an image for my Hero header and used media queries so they will adapt depending on the device the page is being viewed from.
I had at first used a different approach to the JS accordion, but after Maks' lecture on Wednesday I saw that he had a completely different solution - so I changed my function to be more similar to what he showed us.
Not a bad thing since I ended up learning different approaches and Maks' solution ended up being easier to style in CSS - but just as a reference for the future: is it okay if we have a different approach to solve some task than what the coaches show us?
For the rest, I really enjoyed building this page and seeing a tiny bit of what happens behind the scenes (since we used HTTPbin), and look forward to going deeper in the JavaScript~~


code/style.css
```diff
@@ -1,7 +1,352 @@
...
+button { /* styles submit button */
+  margin-top: 25px;
+  font-family: "Alata", sans-serif;
+  font-size: 20px;
+  background-color: #00917c;
+  color: white;
+  padding: 10px;
+  border: none;
+  cursor: pointer;
+}
```
puj, line 179: Remember, if you have ANY other button on your page, this will be the style it receives.  That could have unintended consequences.  I would like to see this as a class.

code/style.css
```diff
@@ -1,7 +1,352 @@
...
+  margin: 0;
+  color: white;
+  font-size: 16px;
+  font-weight: bold;
+  padding: 10px;
+  cursor: pointer;
+}
+
+.button::after {
+  content: "\002B"; /* Created + sign - Unicode character for "plus" sign (+) */
```
puj, line 215: Coooooooool!

code/style.css
```diff
@@ -1,7 +1,352 @@
...
+
+.button {
+  background-color: #00917c;
+  margin: 0;
+  color: white;
+  font-size: 16px;
+  font-weight: bold;
+  padding: 10px;
+  cursor: pointer;
+}
```
puj, line 212: When reading this, without seeing the site, one might ask themselves: Does `.button` apply to all buttons on the whole page?  What kind of button is this?  Where does it appear on the page?

puj (approved):
Hey Vanessa!
 Great job on this project.  The results turned out very nicely!  I think you should be most proud of the custom radio buttons.  That required a lot of effort and exploration I'd imagine.  I'm glad you are not shying away from more advanced topics.  I would like to see experiment more outside the bounds of the lectures with regards to JS, but you will have plenty of time for that.

- Overall, good code structure/formatting
- Really nicely done from responsive standpoint.  The layout makes sense in each case.
- Your commit messages are quite descriptive, continue to refine this.  Very valuable
- "Is it okay if we have a different approach to solve some task than what the coaches show us?"  Absolutely. There are always many ways to solve a problem in development.  Find the best tool for the job if you can.  Exercise a new muscle.  Many solutions bring many learnings.

PR #48, opened 2020-09-04
Project 3 business site - Sanna Dolck Wall

sannadw:
I did the blue level and some things from the other levels. I really had a hard time with the js on the accordion and feel like I have to study a bit more on that to fully understand it.


code/script.js
```diff
@@ -1,9 +1,30 @@
...
+        const accordionItemBody = accordionItemHeader.nextElementSibling;
+        if(accordionItemHeader.classList.contains("active")) {
+            accordionItemBody.style.maxHeight = accordionItemBody.scrollHeight + "px";
+        }
+        else {
+            accordionItemBody.style.maxHeight = 0;
+        }
+    });
+});
+
```
puj, line 16: I'm really happy to see the `.forEach` method here.  Imagine if you had 100 accordion items.  This way you wouldn't need 100 copies of the same code, this would be enough! 👍 

Sometimes it's good to add some comments if you don't feel 100% certain of what's going on:

```javascript
// Get all of the accordion items
const accordionItemHeaders = document.querySelectorAll(".accordion-item-header");

// Iterate through each accordion item
accordionItemHeaders.forEach(accordionItemHeader => {

    // Assign an on click listener to each item
    accordionItemHeader.addEventListener("click", event => {
        // Either add or remove the "active" class from the item when clicked
        accordionItemHeader.classList.toggle("active"); 

        // Set the height of the body so the animation activates 
        const accordionItemBody = accordionItemHeader.nextElementSibling;
        if(accordionItemHeader.classList.contains("active")) {
            accordionItemBody.style.maxHeight = accordionItemBody.scrollHeight + "px";
        }
        else {
            accordionItemBody.style.maxHeight = 0;
        }
    });
});
```

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+  background-repeat: no-repeat;
+  background-position: center center;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.header-background-image h1 {
+  color: #13443b;
```
puj, line 42: Great that you are using the hex code colors!

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+  margin-top: -100px;
+  text-align: center;
+  font-size: 32px;
+}
+
+.header-background-image h2 {
+  color: #7e947f;
+  margin-top: -10px;
```
puj, line 50: Negative margins can usually be achieved in another way.  Try removing your negative margins and setting 
`line-height: 18px` on your `h1` above.

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+  background-position: center center;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.header-background-image h1 {
+  color: #13443b;
+  margin-top: -100px;
```
puj, line 43: Maybe you can make the `header` a flexbox to control the positioning better?  Positive margin on the bottom is better than negative margin on the top.  
Usually always avoid negative css values on margins and padding.

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+    border: 1px solid red;
+  }
+
+/* ACCORDION */
+
+.accordion {
+  padding: 20px;
+  margin: 20px;
+  background-color: rgb(177, 179, 177);
+  box-shadow: 4px 4px #c9c9c9;
```
puj, line 121: Here we are mixing `rgb()` and hex.  Consistency is preferred 👍

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+.accordion-item-header::after {
+  content: "\002B";
+  font-size: 20px;
+  position: absolute;
+  right: 20px;
+}
+
+.accordion-item-header.active::after {
+  content: "\2212";
+}
```
puj, line 152: Very cool with the unicode `content`

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+
+.accordion-item-body {
+  background-color: white;
+  margin-bottom: 15px;
+  max-height: 0;
+  overflow: hidden;
+  transition: max-height 0.2s ease-out;
```
puj, line 159: I like your transition choice here.  Very smooth!

code/style.css
```diff
@@ -1,7 +1,227 @@
...
+} 
+
+.accordion-item-body-content {
+  padding: 20px;
+  border-bottom: 3px solid;
+  border-image: linear-gradient(to right, transparent, #13443b, transparent) 1;
```
puj, line 165: This is such a fun function to play with! Great job!

puj (changes requested):
Hey Sanna!

Great job on this project!  You used a lot of really cool CSS techniques!  And the .foreach usage in the javascript was fantastic.  We'll be revisiting foreach quite a bit in the future.  Your site hero looks great on mobile and tablet.  Cool selection of form inputs and nice accordion transitions!

Changes Requested:
 - Please update your code to be consistently formatted (indentation, etc) in your HTML and CSS files
 - Please update your code with accessible labels in the beginning of your form

PR #50, opened 2020-09-04
Business site 

LillanEriksson:
A business site with forms and a javascript accordion-faq.


code/index.html
```diff
@@ -1,30 +1,107 @@
...
-        <p>Lorem ipsum...</p>
-      </div>
-
-      <div class="question" id="section2">Section 2</div>
-      <div class="answer">
-        <p>Lorem ipsum...</p>
-      </div>
-    </section>
+    <title>Rent-a-classic-VW</title>
+    <link rel="stylesheet" href="style.css">
```
EmOlsz, line 8: let's not forget about `./` in path indicating that we are looking for file `style.css` inside current directory

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Montserrat:wght@900&family=Raleway&display=swap" rel="stylesheet">
```
EmOlsz, line 9: Very nice!

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+</head>
+
+<body>
+    <!-- video or image as a header is cool :) went for image. Cool enough :)-->
+    <div class="hero-image">
```
EmOlsz, line 14: `hero-image` class name says that we should probably expect only image over here, but instead we also have some text nested. From semantical point of view it would be better do name this class something like `hero-container`, don't you think?

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+    <main>
+        <!-- Signup form -->
+        <section class="booking">
+            <div>
+                <h2>Booking</h2>
+            </div>
+            <form class="form" action="http://httpbin.org/anything" method="POST">
+                <div class="form-group">
+                    <label for="name">Name:</label>
+                    <input id="name" type="text" name="name" placeholder="Your full name here" name="nameValue" required>
```
EmOlsz, line 33: Great job, very clean code + great class naming! Nevertheless, consider these changes :
1. Your attributes could be moved to separate lines since there are so many of them (same goes for inputs below)
2. You duplicated `name` attribute over here

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+                <div class="form-group">
+                    <label for="email">E-mail:</label>
+                    <input id="email" type="email" name="email" placeholder="Your e-mail adress here" required>
+                </div>
+                <div class="form-group">
+                    <label for="car">VW Baywindow Bus</label>
+                    <input id="car" type="radio" name="car" value="vwBaywindowBus"></div>
+                <div class="form-group">
+                    <label for="car">VW Beetle</label>
+                    <input id="car" type="radio" name="car" value="vwBeetle">
```
EmOlsz, line 44: To be treated as one radio group, each radio input must share same `name` attribute, however they don't have to (and I even must not) share same `id` - it's unique identifier which should never be duplicated

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+                </div>
+                <div class="form-group">
+                    <label for="numberofadults">How many adults?</label>
+                    <input id="numberOfAdults" type="number" name="numberofadults" min="0" max="3"></div>
+                <div class="form-group">
+                    <label for="numberofkids">How many kids?</label>
+                    <input id="numberOfKids" type="number" name="numberofkids" min="0" max="2">
+                </div>
```
EmOlsz, line 52: Great job, but where are checkboxes we asked you for?

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+                <div class="form-group"><button id="submit" type="submit" name="submit">Submit</button></div>
```
EmOlsz, line 53: In `submit` button, we don't need `name` attribute since it's not the same part of the form. If you will check out HTTPBin site, you will see that your `submit` button slipped into the data inside `form` property

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+        </section>
+
+        <section class="accordion">
+            <!-- Accordion -->
+            <h1>FAQ</h1>
+
+            <div class="question" id="section1">
+
+                <h3>What do we do if we have a breakdown?
+                </h3>
```
EmOlsz, line 64: Indentation could be better over here

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+            </div>
+            <div class="answer">
+                <p>They are old cars, you WILL have breakdowns. Call us first, it's rarely something advanced. If we can't fix it then we call Assistanskåren.</p>
+            </div>
+
+            <div id="section2" class="question">
```
EmOlsz, line 70: It's a super small issue, but one question above your order of attributes is `class` and then `id`, but over here it's opposite

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+
+    </main>
+
+
+
     <!-- dont forget to link the javascript file here! -->
-  </body>
-</html>
+
+    <script src="./script.js"></script>
```
EmOlsz, line 104: Happy to see this tag over here! :)

code/script.js
```diff
@@ -2,8 +2,11 @@
...
-  this.classList.toggle("active")
+    this.classList.toggle("active");
 }
 
 // Selects an HTML element, and calls a function which will be executed when the element is clicked.
-document.getElementById("section1").onclick = toggle
+document.getElementById("section1").onclick = toggle;
+document.getElementById("section2").onclick = toggle;
+document.getElementById("section3").onclick = toggle;
+document.getElementById("section4").onclick = toggle;
```
EmOlsz, line 12: Great job, but we could done it in some loop to make it universal, what do you think?

code/style.css
```diff
@@ -2,6 +2,164 @@
...
+html,
+body {
+    height: 100%;
+    background-color: #c9c5c2;
 }
+
+
+/* The accordion when it's expanded*/
+
+.active+.answer {
```
EmOlsz, line 14: This is completely valid CSS, it would look far more aesthetic written like this : `.active + .answer`

code/index.html
```diff
@@ -1,30 +1,107 @@
...
+        <div class="hero-text">
+            <h1>RENT A CLASSIC VW</h1>
+            <p>Dreaming of those perfect insta-moments? Curious about the vanlife- or vw-community?</p>
+            <p>Rent one of our cars and make YOUR dreams come true.
+            </p>
+        </div>
+    </div>
+
+
+    <main>
```
EmOlsz, line 24: Semantic `main` tag, happy to see it :)

code/style.css
```diff
@@ -2,6 +2,164 @@
...
+
+.hero-text {
+    display: flex;
+    flex-direction: column;
+    color: #c9c5c2;
+    text-align: center;
+    position: absolute;
+    top: 25%;
+    left: 20%;
+    right: 20%;
```
EmOlsz, line 60: Clever and well done!

code/style.css
```diff
@@ -2,6 +2,164 @@
...
+    background-color: #74614d;
+    border-radius: 12px;
+    margin: 20px;
+    box-shadow: 10px 10px 5px #322015;
+}
+
+.form {
+    display: flex;
+    flex-wrap: wrap;
+    flex-direction: column;
```
EmOlsz, line 77: Super 👍

code/style.css
```diff
@@ -2,6 +2,164 @@
...
+@media (min-width: 1200px) {
+    .question,
+    .answer {
+        width: 500px;
+    }
+    .booking div,
+    h1 {
+        padding-left: 100px;
+    }
+}
```
EmOlsz, line 165: And what about responsive hero section?

EmOlsz (changes requested):
Great job Lillemor, you made well project! Unfortunately I will need to ask you to do one change in the code - you haven't met requirement about responsive hero image. Consider using `picture` tag for that, all info about it can be found on Canvas. However, I want to say that that was smallest of our requirements, everything else is done and works great :)

Here are the things you can continue doing this way since you did great job:
1. You use well, semantic HTML tags
2. Class naming is nice (apart from one comment about `hero-image`)
3. You wrote well comments that describes your code
4. Design is nice!
5. I like that you used some form validation
6. Text on hero image was positioned in a clever way

Things you can work on a bit :
1. Name attributes - sometimes it seems that you don't understand them 100%
2. There could be more classes, you seem to style a lot of elements by HTML selectors
3. JS could be universalised a bit

code/script.js
LillanEriksson: I really did not have a clue what I was doing here with js, but it worked. Hopefully I will learn better ways to solve these things the weeks to come.

code/index.html
LillanEriksson: As from the project brief I understood it like we could choose to use all of the input types, or just some of them. I will of course put in a checkbox if it's mandatory.

code/index.html
LillanEriksson: I'm trying to do like you say, but when I save in VSCode it formats it like this.

EmOlsz: Ahh, I see - it varies between local setup of VS Code formatter. You can always install something called `prettier`, it's a formatter for HTML, CSS and JS files, but it's not mandatory

code/index.html
EmOlsz: Yes, this is rather friendly reminder, it's not requirement to pass this project :) Only requirement is present in general comment on top and it's about hero image

code/script.js
EmOlsz: I see, crossing fingers for that! For future if you don't understand any concept, please consider asking questions on Stack Overflow!

code/index.html
LillanEriksson: I had prettier installed, and it madeit look like this. Uninstalled prettier, now I can make it look like this. I think th heroimage should be responsive now.

EmOlsz (approved):

PR #53, opened 2020-09-04
Project 3 - Business site

carolinebirgersson:
I made a business site for the Swim baths of Stockholm stad.


code/style.css
```diff
@@ -2,6 +2,466 @@
...
+  }
+}
+
+/* For mobile */
+@media (max-width: 768px) {
+
+  h1.header-heading {
+    font-size: 48px;
+    width: 75%;
+  }
```
HIPPIEKICK, line 357: Looks like you need an even smaller font-size:

Skärmavbild 2020-09-08 kl  17 04 20

code/index.html
```diff
@@ -3,28 +3,275 @@
...
+          border-radius: 25px;
+          cursor: pointer;
+        ">
+       </textarea>
+      </div>
+    
+      <div class="form-group">
+        <label for="godkännande">Jag har läst igenom och godkänner villkoren.*
+        <input 
+          id="godkännande"
```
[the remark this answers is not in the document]
HIPPIEKICK: Also, the id is mostly used for JavaScript. Use classes for CSS

code/index.html
```diff
@@ -3,28 +3,275 @@
...
+      
+      <div class="form-group">
+        <label for="kurs">Konstsim
+        <input 
+          id="kurs3" 
+          type="radio" 
+          name="courseValue"
+        />
+        </label>
       </div>
```
HIPPIEKICK, line 202: Your inputs' arguments are differing. They should all have a `value` and a `name`. This creates a little bug (it's possible to check more than one radio button). Update this so that all of them have different values (the first one could have the value crawl so it reflects the user's choice), and the name could stay as `courseValue`. A nice feature when working with radio buttons in to make the first radio button checked as default, like this:

```
<input 
  checked
  id="kurs1" 
  type="radio" 
  value="crawl"
  name="courseValue"
/>
```

HIPPIEKICK (approved):
Good job Caroline! It looks super nice with the textarea, and the whole page feels very uniform with the soft colors and the border-radius. Except for the comments I made in the code, here are some suggestions for better readability:

- Some padding to your text inputs so the text doesn't start at the edge
- Either having the same font-size for two texts next to each other or a bigger difference between them

Well done with the form and the accordion!

PR #57, opened 2020-09-04
the Studio - Annika Melin

ankimelin:
Fun with a hero video and I tried animations for the first time! The form was easier to style than I thought, the accordion which looked simpler to do was in the end the tough part both understanding the basic JS and the styling. I had issues with a too big video when publising, tried lfs which did not work, could not get rid of it properly - this caused some manic git commits and pushes to the end before it worked out.


code/index.html
```diff
@@ -3,28 +3,84 @@
...
+      <img class="img-phone" src="./yogaphone.jpeg" alt="Yoga picture">
 
-      <div class="question" id="section1">Section 1</div>
-      <div class="answer">
-        <p>Lorem ipsum...</p>
+      <h1>the Studio</h1>
+    </header>
+
+    <main class="content-container">
+      <div class="content-item">
```
HIPPIEKICK, line 27: Do you need this `div` as a wrapper or can you put the styling on the `form` instead? :)

code/index.html
```diff
@@ -3,28 +3,84 @@
...
+        <form action="http://httpbin.org/anything" method="POST">
+          <h2>Sign up!</h2>
+
+          <label for="name">Name</label><br>
+          <input type="text" id="name" name="name" autocomplete="off"> 
+          <label for="e-mail">E-mail</label><br>
+          <input type="text" id="e-mail" name="e-mail" autocomplete="off"><br>
```
HIPPIEKICK, line 34: There's actually a `type` called email! Check this out: https://www.w3schools.com/tags/att_input_type_email.asp

code/index.html
```diff
@@ -3,28 +3,84 @@
...
+      
+          <p>Choose your city</p>
+          <input type="radio" id="Stockholm" name="city" value="Stockholm" checked="checked">
```
HIPPIEKICK, line 37: The attribute `checked` is a boolean and can therefore be written without a value. The property and the value are the same in this case (checked="checked"). When you have many attributes on one HTML element, it's good to split them up on several lines as well, like this:
```
<input
  checked
  type="radio"
  id="Stockholm"
  name="city"
  value="Stockholm">
```

code/index.html
```diff
@@ -3,28 +3,84 @@
...
-        <p>Lorem ipsum...</p>
+      <h1>the Studio</h1>
+    </header>
+
+    <main class="content-container">
+      <div class="content-item">
+        <form action="http://httpbin.org/anything" method="POST">
+          <h2>Sign up!</h2>
+
+          <label for="name">Name</label><br>
```
HIPPIEKICK, line 31: The element `<br>` is not semantic. Try to work around it by creating other semantic elements and add margin/padding or change the display

code/index.html
```diff
@@ -3,28 +3,84 @@
...
+          </div>
+        </section>
       </div>
-    </section>
-    <!-- dont forget to link the javascript file here! -->
+    </main>
+
+  <footer>
+    
+  </footer>
```
HIPPIEKICK, line 81: Remove unused code!

code/style.css
```diff
@@ -1,7 +1,212 @@
...
+    animation-delay: 2s;
+    animation-direction: alternate;
+    animation-iteration-count: infinite; 
+  }
+}
+
+@keyframes slowyellow {
+  from {color: white;}
+  to {color: yellow;}
+}
```
HIPPIEKICK, line 66: Yey, cool animations Annika!

code/style.css
```diff
@@ -1,7 +1,212 @@
...
+  }
+}
+
+/*** CONTENT: SIGNUP FORM AND ACCORDION ***/
+
+.content-container {
+  display: flex;
+  justify-content: space-evenly;
+  flex-wrap: wrap;
+  padding: 40px 20px 40px 20px;
```
HIPPIEKICK, line 81: When the top and bottom have the same padding, and the left/right have the same - you could pass in two values instead of four (the first one applies to top/bottom and the second one to left/right):
`padding: 40px 20px;`

code/style.css
```diff
@@ -1,7 +1,212 @@
...
+  }
+}
+
+/*** SIGNUP FORM ***/
+
+form {
+  height: 500px;
+  padding: 20px 40px 20px 40px;
+  border-radius: 20px;
+  box-shadow: 0 4px 8px 4px rgba(0, 0, 0, 0.2);
```
HIPPIEKICK, line 122: I'm liking this box-shadow!

code/style.css
```diff
@@ -1,7 +1,212 @@
...
+  width: 100%;
+  margin: 4px 0 8px 0;
+  padding: 8px;
+  border: none;
+  border-radius: 20px;
+  background-color: rgb(240, 238, 238);
+}
+
+input[type=text]:focus, input[type=e-mail]:focus {
+  outline: none;
```
HIPPIEKICK, line 136: Be aware that the `focus` is there for accessibility reasons. We will talk more about accessibility so no worries :)

code/style.css
```diff
@@ -1,7 +1,212 @@
...
+.question {
+  padding-bottom: 8px;
+  font-size: 20px;
+  cursor: pointer;
+}
+
+.answer {
+  display: none;
+  padding-bottom: 8px;
+}
```
HIPPIEKICK, line 197: Would be nice to the eyes if these looked a little more different from each other

HIPPIEKICK (approved):
Good job Annika! Your page is responsive and you managed well with the hero video/images, accordion, and the form. If you have time, challenge yourself and see what other input types there are and what attributes you can play around with (`required` might be useful in this form).

code/index.html
ankimelin: I wanted to have a class making it possible to style both the the form and accordion, so I'll leave it :-)

code/index.html
ankimelin: Great, thanks - makes sense!

code/index.html
ankimelin: That makes the code so much easier to read!

code/index.html
ankimelin: Noted, though this was not a quick fix for me.. so I will leave it for now and have it in mind going forward.

code/index.html
ankimelin: It is styled though, so not completely unused!

code/style.css
ankimelin: Something fun had to happen :-)

code/style.css
ankimelin: Of course - thanks!

code/style.css
ankimelin: Yey!

code/style.css
ankimelin: Is outline-style: none the way to go here to achieve the styling but keep accessibility?

code/style.css
ankimelin: I agree!! Will spend a little time here.

ankimelin:
Thanks again Matilda for a great code feedback!

code/style.css
HIPPIEKICK: Here are some suggestions on how to deal with it: https://www.a11yproject.com/posts/2013-01-25-never-remove-css-outlines/

ankimelin: A late thank you, what a great short text about it!

PR #58, opened 2020-09-04
Period sponsors - business site!

idsintehittapa:
I've created a site witch has
-  a header with responsive image
-  a signup form with at least three different input types
- a FAQ accordion, built with JavaScript

Furthermore, the site has
- text, number, check, radio, email, submit form elements
- validation to the form within the HTML input
- on hover over buttons and checkbox, the mouse cursor changes to something other than the default
- styled checkboxes and radio buttons
- an a slide-down effect animations when the accordion expands and collapses

I've tested the site on desktop with firefox, opera and on different Iphones. The checkbox does not work perfectly on smaller screens.


index.html
```diff
@@ -0,0 +1,267 @@
...
+    </h2>
+    <h3 class="manifesto-h3">
+      / Göteborg's Stadsmission
+    </h3>
+
+    <div class="p-manifesto-wrapper">
+      <p class="p-manifesto">Once a month. Twelve months a year. For many, it is everyday-life to buy a package of pads,
+        disinfect a menstrual cup or throw a tampon in the bag. But not for everyone </p>
+    </div>
+  </section>
```
HIPPIEKICK, line 49: It's really effectual working with spans and different font-sizes but have in mind that it affects the readability

index.html
```diff
@@ -0,0 +1,267 @@
...
+
+
+  <!-- ACCORDION-->
+  <section class="accordion" id="accordion">
+
+    <div>
+      <h1 class="accordion-question">I have questions about the project!</h1>
+    </div>
```
HIPPIEKICK, line 57: Does this need to be wrapped in a `div`?

index.html
```diff
@@ -0,0 +1,267 @@
...
+  </section>
+
+
+  <!-- Signup form -->
+
+  <section class="the-form" id="the-form">
+
+    <fieldset class="field-container">
+      <legend>
```
HIPPIEKICK, line 121: 🥳

index.html
```diff
@@ -0,0 +1,267 @@
...
+          name="civicNrValue" 
+          required>
+          *
+        </div>
+
+        <div class="person">
+          <label for="firstName">
+            <p class="p-form-small">First name</p>
+          </label>
+          <input id="firstName" type="text" name="firstNameValue" required>
```
HIPPIEKICK, line 169: Like you did with the last input, it's a good idea to split the arguments to different lines if they are many:
```
<input
  id="firstName"
  type="text"
  name="firstNameValue"
  required>
```

index.html
```diff
@@ -0,0 +1,267 @@
...
+        A package of pads or tampons costs approximately 28 SEK .
+        You decide how many packages you want and can contribute with. </p>
+
+      <form class=”form” action="http://httpbin.org/anything" method=”POST”>
+        <p class="p-form">SELECT AN AMOUNT</p>
+
+        <div class="amount-container">
+          <div class="amount">
+            <label for="SekValue" class="radio-button-container">
+              <p class="p-form-small">28 sek/month</p>
```
HIPPIEKICK, line 136: Style the label instead of nesting a `p` tag

script.js
```diff
@@ -0,0 +1,72 @@
...
+//function toggle() {
+  //this.classList.toggle("active")
+//}
+
+// Selects an HTML element, and calls a function which will be executed when the element is clicked.
+//document.getElementById("section1").onclick = toggle
+//document.getElementById("section2").onclick = toggle
+//document.getElementById("section3").onclick = toggle
+//document.getElementById("section4").onclick = toggle
+//document.getElementById("section5").onclick = toggle
```
HIPPIEKICK, line 13: Remove unused code

script.js
```diff
@@ -0,0 +1,72 @@
...
+section5.onclick = () => {
+  const answer = document.getElementById("answer5");
+  answer.classList.toggle("active-image");
+
+  document.getElementById("answer1").classList.remove('active');
+  document.getElementById("answer2").classList.remove('active');
+  document.getElementById("answer3").classList.remove('active');
+  document.getElementById("answer4").classList.remove('active');
+}
+
```
HIPPIEKICK, line 72: Since this code looks almost the same for all sections, it would be cool if we can shorten it without repeating ourselves. If you want, take a look at Maksymilian's notes from Week 3 lecture 2 (where he does a forEach loop) :)

style.css
```diff
@@ -0,0 +1,522 @@
...
+this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
+
+@import url('https://fonts.googleapis.com/css2?family=Monoton&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Viga&display=swap');
+@import url('https://fonts.googleapis.com/css2?family=Montserrat:wght@300&display=swap');
+
+*, *::before, *::after {box-sizing: border-box}
+
+
+body{
```
HIPPIEKICK, line 12: A really small detail, but there should be a space between the elements and the opening curly bracket

style.css
```diff
@@ -0,0 +1,522 @@
...
+  border-radius: 50%;
+}
+
+.radio-button-container:hover input ~ .custom-radio{
+  background-color: #444444;
+}
+
+.radio-button-container input:checked ~ .custom-radio{
+  background-color: #444444
+}
```
HIPPIEKICK, line 354: Consider having the hover effect lighter than the checked one to clarify the difference

style.css
```diff
@@ -0,0 +1,522 @@
...
+  font-size: 16px;
+  user-select:none;
+}
+
+.checkbox-container input{
+  position: absolute;
+  opacity: 0;
+  cursor: pointer;
+  height: 17;
+  width: 17;
```
HIPPIEKICK, line 383: Specify the unit

style.css
```diff
@@ -0,0 +1,522 @@
...
+input[type = "number"]{
+  font-family: 'Montserrat', sans-serif;
+  font-size: 0.8em;
+  height: 100px;
+}
+
+input[type = "submit"]{
+  font-family: 'Montserrat', sans-serif;
+  font-size: 0.8em;
+}
```
HIPPIEKICK, line 443: When you have the same styling for two elements, you could just separate the elements with a comma:
```
input[type="text"], input[type="submit"] {
  font-family: 'Montserrat', sans-serif;
  font-size: 0.8em;
}
```

HIPPIEKICK (approved):
You managed well with the responsiveness - I like that you're coding mobile-first and are keeping the content of the media queries to a minimum. It looks good on all widths.

It gets a little bit messy visually with all different font sizes and red spans. I want you to take a look at the amount of HTML-elements, your class names and id's as well. This is because you don't need as many as you think you need - if that makes sense. You only need to specify an id if you want to access it in JavaScript. You only need to specify different class names if you're going to style elements differently from each other.

Good job with the sliding accordion and super nice with the styled checkboxes and radio buttons 🥳

PR #60, opened 2020-09-04
Project 3-Jessica Pandtha- Businesssite

jespa374:
I tested the responsiveness in the different breakpoints in Safari and Chrome and they worked as I expected. Also tested in my iPhone and it worked and looked the same.


code/index.html
```diff
@@ -3,28 +3,82 @@
...
-
-      <div class="question" id="section2">Section 2</div>
-      <div class="answer">
-        <p>Lorem ipsum...</p>
+    </header>
+    <main>
+      <h2 class="wine-gallery-heading">Our wines</h2>
+      <section class="wine-gallery">
+        <div>
+          <img src="Images/Red-wine.jpg" alt="Red wine">
```
EmOlsz, line 26: folders inside your project should be written in lowercase

code/index.html
```diff
@@ -3,28 +3,82 @@
...
+        </div>
+        <div>
+          <img src="Images/Sparkling-wine.jpg"alt="Sparkling wine">
+        </div>
+      </section>
+      <!-- Signup form -->
+      <div class="form-container">
+        <form action="https://httpbin.org/anything" method="POST" class="sign-up-form">
+          <h2>Sign up</h2>
+          <label>Name</label>
```
EmOlsz, line 42: Your labels are not connected to exact inputs, do you see? To do so, you should either nest inputs inside labels, or add `id` attribute for input and `for` attribute with the same value to label

code/index.html
```diff
@@ -3,28 +3,82 @@
...
+          <input type="text" name="name" class="sign-up-info"required>
+          <label>E-mail</label>
+          <input type="email" name="email" class="sign-up-info" required>
+          <label>Telephone</label>
+          <input type="number" name="telephone" class="sign-up-info">
+          <label>Would you like to subscribe to our newsletter?</label>
+          <label><input type="radio" name="subscribe">Yes</label>
```
EmOlsz, line 49: Look, and here you've done it :)

code/index.html
```diff
@@ -3,28 +3,82 @@
...
+          <label><input type="radio" name="subscribe">No</label>
+          <label><input type="checkbox"name="conditions">Accept all terms and conditions</label>
+          
+          <button class="submit-button">Submit</button>
```
EmOlsz, line 53: You are missing `type="submit"` attribute over here

code/index.html
```diff
@@ -3,28 +3,82 @@
...
+          </div>
+        </section>
       </div>
-    </section>
-    <!-- dont forget to link the javascript file here! -->
+    </main>
+    <footer class="footer" id="footer">
+      <h4>Contact information</h4>
+      <p>Telephone: +46 70 89 99 99</p>
+      <p><Address>
```
EmOlsz, line 76: 1. Since `address` tag is nested inside `p` tag, let's do proper formatting and do a new line for `address`
2. `address` tag should be written lowercase

code/style.css
```diff
@@ -1,7 +1,231 @@
...
+  height: 25px;
+}
+
+.submit-button {
+  width: 70px;
+}
+
+input::-webkit-outer-spin-button,
+input::-webkit-inner-spin-button {
+  -webkit-appearance: none;
```
EmOlsz, line 62: Very nice, you did some research I see :)

code/style.css
```diff
@@ -1,7 +1,231 @@
...
+
+.wine-gallery img {
+  width: 100%;
+  margin-bottom: 0;
+}
+
+}
+
+
+
```
EmOlsz, line 179: We don't need so many blank lines over here, let's delete them

code/style.css
```diff
@@ -1,7 +1,231 @@
...
+   justify-content: space-between;
+  }
+
+.sign-up-form {
+  width: 50%;
+  }
+
+.accordion {
+  width: 50%;
+  }
```
EmOlsz, line 219: from now until the end there is some wrong formatting, let's fix that as well

code/index.html
```diff
@@ -3,28 +3,82 @@
...
+          <h2>Sign up</h2>
+          <label>Name</label>
+          <input type="text" name="name" class="sign-up-info"required>
+          <label>E-mail</label>
+          <input type="email" name="email" class="sign-up-info" required>
+          <label>Telephone</label>
+          <input type="number" name="telephone" class="sign-up-info">
+          <label>Would you like to subscribe to our newsletter?</label>
+          <label><input type="radio" name="subscribe">Yes</label>
+          <label><input type="radio" name="subscribe">No</label>
```
EmOlsz, line 50: The only problem is that your inputs doesn't have `value` attribute and because of that checking both Yes or No answers are saved in form as `subscribe: on`

EmOlsz (approved):
Very nice project Jessica, congratulations! I'm especially happy to see BEM methodology and validation added to you form :) You seem to have solid knowledge on the fundaments of HTML. On the other hand, sometimes you are making mistakes when it comes to choosing uppercase naming. You could also practice `label`s in forms so you remember about `type="submit"` attribute and `for`-`id` pair for `label`-`input`.
Apart from that, great job :)

PR #63, opened 2020-09-04
Project 3 - Free walking tour

Corneliaryas:
I tried to make it mobile-first and to do the layout with CSS Grid (since I didn't do Grid last week), which took up a lot of the time. Still need to practice that a lot more before I get it.  Fun to style the form and accordion! Would have loved to be able to put more time on that and the validation. 

Still struggling with when to use fixed values of height and width and when not to.


code/index.html
```diff
@@ -3,28 +3,93 @@
...
+    </header>
 
-    <!-- Signup form -->
+    <main>
+      <section class="intro">
+        <h2>Join us on a tour around Söder</h2>
+        <p>The tour will take you around Södermalm in Stockholm and show you the most precious and hidden gems of the island. 
+        We will both walk you through history and tell you the most up to date news. 
+        We'll meet up at Slussen, take you around, and return you safely to slussen again.  </p>
+      </section>
```
Cilpersson, line 28: Great semantics with the header-tags and nicely indented 🤩

code/index.html
```diff
@@ -3,28 +3,93 @@
...
-      <h1>FAQ</h1>
+      <section class="sign-up">
+        <h2>Sign up here</h2>
+   
 
-      <div class="question" id="section1">Section 1</div>
-      <div class="answer">
-        <p>Lorem ipsum...</p>
-      </div>
+        <form action="https://httpbin.org/anything" method="POST">
```
Cilpersson, line 34: Perfect 👍

code/style.css
```diff
@@ -1,7 +1,384 @@
...
+
+/*** MAIN ***/
+
+main {
+  display: grid;
+  grid-template-rows: repeat(4, 1fr);
+  align-items: center;
+  justify-content: center;
+  
+}
```
Cilpersson, line 63: I love to see that you have given grid a try, mastering both grid and flexbox will give you endless styling possibilities. Great job 🏅

code/style.css
```diff
@@ -1,7 +1,384 @@
...
+
+.radio-container .custom-radio:after {
+  top: 6px;
+  left: 6px;
+  width: 12px;
+  height: 12px;
+  border-radius: 50%;
+  background-color: lightsalmon;
+}
+
```
Cilpersson, line 197: B 👏 R 👏 A 👏 V 👏 O

code/style.css
```diff
@@ -1,7 +1,384 @@
...
+  cursor: pointer;
+}
+
+.answer {
+  margin: 10px 0;
+  background: white;
+  height: 0;
+  overflow: hidden;
+  color: white;
+  transition: height 0.5s;
```
Cilpersson, line 238: Nice with the transition!

code/style.css
```diff
@@ -1,7 +1,384 @@
...
+  font-weight: bold;
+  font-size: inherit;
+  font-family: inherit;
+}
+
+/*** FAQ ***/
+
+.accordion {
+  width: 100%;
+  height: 300px;
```
Cilpersson, line 215: Regarding when to use and not use fixed widths/heights. Here's a good example, when one of your .question-containers are active and the answer is shown it expands outside of the .accordion-containers fixed height. 
Screenshot 2020-09-10 at 15 29 32

This only happens in mobile view however. So don't use fixed heights or widths if the size of the containers content might change. An option here would be to use `min-height: 300px;` then the container will never be less than 300px in height, but can expand if the content of it does. This is tricky and the beginning but since you asked about it I thought I should give you a thorough answer.

Cilpersson (approved):
Cornelia, you have created something you should be very proud of. First and foremost the code nicely indented through all files (this is the most important part and you have really well structured code 🎖),  but it also looks great! 

I have written a comment regarding fixed heights/widths that I hope will bring some clarity to you, if not let me know and I'll try to explain further. Besides that there wasn't a lot to comment about, because you have pushed yourself hard and it really shows. So the only thing I would love to see for the next coming projects would be that you commit to git even more, you messages are good but try to commit at least 3-4 times per coding session. Another thing I would suggest is to remove images-files you're not using (such as the -png.files in the root-folder). But now I'm just nitpicking 🤪

Again, great job 🥳🎉

PR #105, opened 2021-01-29
new pull request

Sartish:


code/script.js
```diff
@@ -1,9 +1,51 @@
...
+
+//1 get the icon ids' and store as variables
+const minusElement1 = document.querySelector('#minus1')
+const plusElement1 = document.querySelector('#plus1')
+
+const minusElement2 = document.querySelector('#minus2')
+const plusElement2 = document.querySelector('#plus2')
+
+const minusElement3 = document.querySelector('#minus3')
+const plusElement3 = document.querySelector('#plus3')
```
HIPPIEKICK, line 14: Nice! This could also be done by CSS styling, so you use the "active" class and then you can style with ::after or ::before and the CSS property `content` :)

code/script.js
```diff
@@ -1,9 +1,51 @@
...
+ this.classList.toggle("active")
+ minusElement2.classList.toggle('show')
+ plusElement2.classList.toggle('hide')
+ }
+
+ function toggleThree() {
   this.classList.toggle("active")
+  minusElement3.classList.toggle('show')
+  plusElement3.classList.toggle('hide')
+  }
```
HIPPIEKICK, line 36: Since the functions are doing the same thing but to different elements, think about how we can revise this to only have one function but use it many times on the different variables

code/script.js
```diff
@@ -1,9 +1,51 @@
...
+ 
+
+function clickAlert() {
+  alert("Welcome to Julia's🙋‍♀️. View your inbox for a starting treat👀🍲")
 }
```
HIPPIEKICK, line 41: Nice add 🥳

HIPPIEKICK (changes requested):
Good job Sara! I like how aligned your design is, looks super! My main comments are about not forgetting the `label`s as they are important for accessibility reasons, and the `value` and `name` attributes of the radio buttons and checkboxes. Also, please check all your attributes, so that they look like this:
`<div class="header-title">`
and not like this:
`<div class = header-title>`
(all values should have quotation marks, and there should be no space around the `=` ).

Check out the [general feedback document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475) to read more :)

code/index.html
```diff
@@ -2,29 +2,185 @@
...
+        </div>
+        <div class="name-fields">
+          <input name="first-name" type="text" placeholder="First name" required>
+          <input name="last-name" type="text" placeholder="Last name" required>
+        </div>
+
+        <div class="email-password">
+          <input name ="email" type="email" placeholder="Email adress" required>
+          <input name ="password" type="text" placeholder="Password" required>
+        </div>
```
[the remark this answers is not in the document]
Sartish: Ah yes! it is because I thought it lookes nicer without labels 😄 Like only have it in the placeholder😅

code/script.js
Sartish: Ah nice tip, didn't think about this 🤩

HIPPIEKICK:
Hi again! You still need to change your attributes' values (see previous comment about that). Regarding the labels, you don't have to add text to them, but they are good for accessibility reasons because they allow a user to click the label themselves to place the cursor inside the respective input box :)

KarolCislowski:
Hello, it's your project. Do you remember me? I may be imperfect, but I believe you can still change that before you sit down to your final masterpiece!
image

Sartish:
Hi! 
Thank you for the reminder, I have added labels also now but without text :) And name attribute to check-boxes :)

KarolCislowski (approved):
Nice page 👍🏻

PR #107, opened 2021-01-29
Project 3 - Sofia 

Sofarz:
Finished the project. Did some of the red stretch goals by adding validation to the form and some styling to the accordion. 

Decided to use the JavaScript method that Maksimilian showed us during the code sessions, so I struggled a little with understanding the other method that was presented.


HIPPIEKICK (approved):
Good job Sofia! You seem to have grasped the form and the JavaScript well, my main comment is to try to scale down your HTML content (cutting off some `div`s). It's nothing you have to change on this project but try to think about it in the future so we don't end up with a lot more HTML than necessary. Check out the [general feedback document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475) as well, and the [code guidelines](https://www.notion.so/technigo/Guidelines-for-how-to-write-good-code-69ffb59eac4e4358a7c5752efb873c4e) if you want to recap.

PR #119, opened 2021-01-30
Project 3 - Mette

ettem:
It has been a hard week and I've had difficulties pulling the project through, so I'm glad if I managed to fix the minimum requirements this week!

Responsive layout is still a struggle so I've spent endless hours trying to fix that including using a video instead of an image for header. Tried to use different size for the video but it didn't work, still somehow the video looks responsive to me now!

Also tried to style the form but since I didn't have so much time left I couldn't style the radio buttons and the checkbox, googled a bit but it seems quite difficult.

One thing I wonder is about the white space (margin) between the header and the main section. Tried to get rid of it but it's still there as you can see...


code/index.html
```diff
@@ -1,30 +1,167 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@100&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=Roboto:ital,wght@0,100;1,400&display=swap" rel="stylesheet">
   </head>
+
   <body>
-    <!-- video or image as a header is cool :) -->
+     <!-- video or image as a header is cool :) -->
+    <header class="header">
+      <div class="hero">
+        <video width="" height="" playsinline autoplay muted loop poster="" class="bgvideo">
```
HIPPIEKICK, line 20: You can omit the `width` and `height` attributes if you don't want to give them values

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+        <h1>A Forgotten Craftsmanship</h1>
+        <p>Learn Bookbinding Today!</p>
+      </section>
+    </header>
 
-      <div class="question" id="section1">Section 1</div>
-      <div class="answer">
-        <p>Lorem ipsum...</p>
-      </div>
+    <div class="main">
```
HIPPIEKICK, line 48: Here we can use the `main` element for a more semantical approach

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+
+      <!-- Signup form -->
+      <section class="form">
+        <form action="http://httpbin.org/anything" method="POST">
```
HIPPIEKICK, line 52: Do you need the `section` or can you style the `form` directly?

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+        <h2>Sign up here!</h2>
+        <h5>To get more information and notification alernatives.</h5>
+     
+        <div class= "text-input">
```
HIPPIEKICK, line 56: An errant space 👀

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+          <label class="name">
+            Name<br>
+            <input class="textfield" type="text" name="message" required></input><br>
```
HIPPIEKICK, line 59: Inputs are self-closing tags because they don't have any nested content, so you can write like this. And maybe it makes more sense if the `name` attribute was "name"? :)
```
<input
  required
  class="textfield"
  type="text"
  name="name"
/>
```

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+          <label class="e-mail">
+            E-mail<br>
+            <input class="textfield" type="email" name="email" required><br>
+          </label>
+        </div>
+
+        <div class="select-menu-input">
+          <label>
+            Course Location:<br>
+            <select name="location" id=""><br>
```
HIPPIEKICK, line 70: No empty values 😇

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+          <label><br>
+            <input type="radio" name="ApplicationPapers">
+            Application Papers
+          </label>
+
+          <label><br>
+            <input type="radio" name="MonthlyNewsletter">
+            Monthly Newsletter
+          </label>
+        </div>
```
HIPPIEKICK, line 99: If the user shoul be able to check more than one option, these should be checkboxes. If the user should only be able to check one, you should make sure to have the same `name` on all of the radio buttons, but different values. Read more in week 3's general feedback document

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+          <div class="accordion-item">
+            <div class="accordion-item-header">
+              <h4>Application</h4>
+            </div>
+            <div class="accordion-item-body">
+              <div class="accordion-item-body-content">
+                <p>You apply...</p>
+              </div>
+            </div>
+          </div>
```
HIPPIEKICK, line 151: Nice accordion! Do you need all these `div`s though? 😇

code/script.js
```diff
@@ -1,9 +1,21 @@
...
 // console.log(this) to see how it works
-function toggle() {
+/*function toggle() {
   this.classList.toggle("active")
 }
 
 // Selects an HTML element, and calls a function which will be executed when the element is clicked.
 document.getElementById("section1").onclick = toggle
+
+My code starts here -->*/
```
HIPPIEKICK, line 11: Remove unused code

code/script.js
```diff
@@ -1,9 +1,21 @@
...
+
+//defining variable
+const accordionItemHeader = document.querySelectorAll(".accordion-item-header");
+
+//function; for every variable toggle the class active when clicking on header
+accordionItemHeader.forEach(accordionItemHeader => {
+    accordionItemHeader.addEventListener("click", event => {
+        accordionItemHeader.classList.toggle("active");
+    });
+});
```
HIPPIEKICK, line 21: Yey, nice solution! 👍 👏

code/style.css
```diff
@@ -2,6 +2,336 @@
...
+  padding: 5px;
+  margin-left: 15%;
+  margin-bottom: 2%;
+  border-radius: 5px;
+  border: none;
+}
+
+input:valid {
+  background-color: lightgreen;
+}
```
HIPPIEKICK, line 133: Nice feature!

code/index.html
```diff
@@ -1,30 +1,167 @@
...
+
+        <div class="submit-button">
+          <button type="submit">Sign up</button>
+        </div>
+
+        </form>
+      </section>
+
+      <!--Accordion-->
+      <section class="QandA">
```
HIPPIEKICK, line 116: Remember kebab-case

HIPPIEKICK (changes requested):
Hi Mette! Regarding styling checkboxes and radio buttons - it's a little bit tricky as you say, but if you take a look at the black level goals in the brief, there are links to some codealongs there if you wanna try it out! :)

I really like your solutions, however there are some small things for you to fix. Take a look at my comments. Check out the [general feedback document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475) for more tips!

code/index.html
ettem: I believe I need it! Removing it will cause the background of the form to disappear and I don't know how to get to it without this section although a bit weird!

HIPPIEKICK (approved):

PR #131, opened 2021-01-31
Week 3 project - Business site

johlofg:
I got some issues with margin, as you can see everything tends to go to the left a bit. And the hero img is wider then sign up and accordion. Just realised that its maybe because its not in the site content div as the other two?


HIPPIEKICK (changes requested):
Hi Johan! I made some comment for you to look over. The comments about labels and indentation need to be fixed. Check out the [general feedback document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475) for further reference :) Have a good weekend!

johlofg:
I think I solved everything now, otherwise I missunderstood you somewhere :) But I maybe use dubble indention in the css file? If so I need to change my default space there I guess.

HIPPIEKICK (approved):

PR #139, opened 2021-01-31
Project 3 / Anna-Kaisa Honkonen

honkoan:
I would have wanted to do the red and black level requirements but I hadn't enough time this week. This was really nice project!


code/index.html
```diff
@@ -1,30 +1,91 @@
...
+          <input name="firstName" type="text" id="firstName" placeholder="Your first name" /><br>
+          <label for="lastName">Last name:</label>
+          <input name="lastName" type="text" id="lastName" placeholder="Your last name" /><br>
+          <label for="email">Email:</label>
+          <input name="email" type="email" id="email" required placeholder="your@email.com" />
+        </div>
+        <div class="radiobuttons">
+          <h3>Which one do you prefer?</h3>
+          <input type="radio" name="bike" value="trek" />
+          <label for="trek">TREK</label>
```
EmOlsz, line 44: labels and inputs can be connected via `for`-`id` pair, not `for`-`value` pair.
In your case, it should be :

```
<input type="radio" id="bike" name="bike" value="trek" />
<label for="bike">TREK</label>
```

Let's fix that here and in other input elements ;)

code/index.html
```diff
@@ -1,30 +1,91 @@
...
+          <input type="radio" name="bike" value="cannondale" />
+          <label for="cannondale">Cannondale</label>
+          <input type="radio" name="bike" value="other" />
+          <label for="other">Other</label>
+        </div><br>
```
EmOlsz, line 49: We shouldn't use `<br />` tags. As a matter of fact, you don't need them here, because you use `<div />` tag, which already has feature to break line. Let's delete them

code/index.html
```diff
@@ -1,30 +1,91 @@
...
+  <meta charset="utf-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Luksanmäki-MTB</title>
+  <link rel="stylesheet" href="style.css">
+</head>
+
+<body>
+
+  <header>
+    <picture>
```
EmOlsz, line 14: Great job :)

code/index.html
```diff
@@ -1,30 +1,91 @@
...
+          <p class="answer" id="answer4">Lorem ipsum...</p>
+        </div>
+      </section>
+    </div>
+
+  </div>
+
+  <footer>Technigo bootcamp spring 2021 project 3.</footer>
+
+  <script src="./script.js"></script>
```
EmOlsz, line 87: Thank you for remembering about `./`

EmOlsz (changes requested):
Hey!
Very nice project, good job! There are only two things I would like you to fix - connection between input and label, and `<br />` tags. You can find detailed explanation in comments.
Moreover, feel free to check out [Week 3 Master document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475)

EmOlsz (approved):

PR #140, opened 2021-01-31
Project 3

Rephili:
Project with a hero image and a form. Doesn't meet all the requirements at the moment, but I plan to add the accordion later on.


HIPPIEKICK:
Hi Rebecca! Can you submit a link to your project on netlify as well? :)

README.md
```diff
@@ -1,12 +1,10 @@
...
 
-Replace this readme with your own information about your project. 
-
-Start by briefly describing the assignment in a sentence or two. Keep it short and to the point.
+Site with a hero image and a form.
 
 ## The problem
 
-Describe how you approached to problem, and what tools and techniques you used to solve it. How did you plan? What technologies did you use? If you had more time, what would be next?
+Due to health circunstances this week I didn't manage to finish the accordion, but I plan to do it as soon as I can (and also add a link to view it live).
```
HIPPIEKICK, line 7: Ahh, I see. No worries, I'll do a code review on the existing code 👍 Hope you're feeling better 🌺

code/index.html
```diff
@@ -4,27 +4,88 @@
...
-  <body>
+    <body>
+
     <!-- video or image as a header is cool :) -->
+      <div class="hero-image-container">
+        <div class="hero-title-container">
+          <h1><span>Pura Vida travels</span></h1>
+          <h2>Tailor-made trips to Costa Rica</h2>
+        </div>
+      </div>
```
HIPPIEKICK, line 17: Maybe this could be a `<header>`? :) I think the span is redundant.

code/index.html
```diff
@@ -4,27 +4,88 @@
...
+      
+    <!-- Wrap accordion and signup with flexbox -->
+
+  <div class="wrapper">
```
HIPPIEKICK, line 21: Maybe this could be a `<main>` for a more semantical approach

code/index.html
```diff
@@ -4,27 +4,88 @@
...
+
+    <!--Sign up -->
+    
+      <section class="contact-form">
+        <form action="https://httpbin.org/anything" method="POST">
+          <h1>Contact us</h1>
+
+          <div class="textfield">
+            <label>Full name</label>
+            <input type="text" name="" value="" placeholder="Write your name">
```
HIPPIEKICK, line 31: When you have the input outside of the label like this, you need to create a connection using `for` and `id` attributes. You can read more about this in the general feedback document for week 3.

code/index.html
```diff
@@ -4,27 +4,88 @@
...
+          </div>
+
+          <div class="textfield">
+            <label>Email</label>
+            <input type="text" name="" value="" placeholder="Write your email">
```
HIPPIEKICK, line 36: There's a type called email that you can use here! And if you don't want a name or value on your inputs - omit them instead of leaving them blank

HIPPIEKICK: Although it's good to give them names to easier be able to read the submission of the form :)

code/index.html
```diff
@@ -4,27 +4,88 @@
...
+          </div>
+
+          <label class="checkbox">
+            <input type="checkbox" checked="checked">
+            <span class="checkmark">Get more info!</span>
+          </label>
+
+          <div class="button">
+            <input type="submit" name="Send">
```
HIPPIEKICK, line 45: The submit could be a button or an input (both of type submit). You can omit the div here.

code/index.html
```diff
@@ -4,27 +4,88 @@
...
+          </div>
+
+          <div class="content">
+            <div class="question4">What packages do you offer?</div>
+            <div class="answer">
+              <p>Bla bla bla.</p>
+            </div>
+          </div>
+
+        </section>
```
HIPPIEKICK, line 81: Looks like you have a good start here! Just remember that class names are for CSS styling so maybe these can have the same class names but different ids (if they are going to have the same styling that is).

code/style.css
```diff
@@ -2,6 +2,171 @@
...
+  margin-top: 24px;
+}
+.wrapper {
+  width: 100%;
+  height: 40%;
+  border: 2px solid black;
+}
+
+.contact-form {
+  width: 30%;
```
HIPPIEKICK, line 64: Let the form take up the whole width on mobile since they are already narrow 😇

HIPPIEKICK (changes requested):
You have a good start Rebecca! I gave you some pointers regarding your form, and use the channels if you need more guidance regarding the accordion. Here's a link to the [general feedback document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475) as well.

PR #146, opened 2021-01-31
Week 3 Project Business Site

sarianne-peltola:
I think I reached Red Level Requirements. I think this was very nice project! I still have some difficulties in creating a well-working and responsive structure with equal margins everywhere. :)


code/style.css
```diff
@@ -1,7 +1,321 @@
...
+  .footer {
+  flex-direction: row;
+}
+
+/** LAPTOP **/
+
+@media (min-width: 992px) {
+
+  .join {
+    float: left;
```
JennieDalgren, line 302: **float** is a bit of an oldschool approach. Not something you want in your code in 2021 for recruiters to see. 
Try to use flexbox instead. (or why not try out grid when (if) you have time!)
🙈

code/style.css
```diff
@@ -1,7 +1,321 @@
...
+    width: 60%;
+    margin: 30px;
+  }
+
+  .faq {
+    float: left;
```
JennieDalgren, line 308: **float** is a bit of an oldschool approach. Not something you want in your code in 2021 for recruiters to see. 
Try to use flexbox instead. (or why not try out grid when (if) you have time!)
🙈

JennieDalgren (approved):
Wow!

Really nice work with this project. I like you small detailed animation, such as the logo.
Great structure of the code files and good use of comments.
Also great work with commits and the commit messages.

If you want to dig deeper into this weeks learning objectives, here's the[ general feedback document](https://www.notion.so/technigo/Week-3-Business-site-9f3ef7b1c0ee480c8a10b9c7e460c475)

Keep up the good work. 💪🏼

PR #163, opened 2021-02-10
Drag Doll, Hannah Jesinkey

jesinkey:
Sorry for the delay! Didn't realise I had missed this. Poyan said it was OK.
/ Hannah


HIPPIEKICK (changes requested):
Hi Hannah!
Some pointers below:
**HTML**
- Instead of writing the`value` for the inputs which will force the user to delete what you've written, you could look into `placeholder`
- Attribute's values needs quotation marks, e.g. `id="fname"` instead of `id=fname`
- Check indentation, and make sure the tags correspond -> `<h3></h3>`
- From requirements: Site should have a signup form with _at least three different input types_

**JS**
- Remove console.logs
- Be consistent whether you're using semicolons or not
- The toggle function that was already declared is reusable thanks to the `this` keyword. You don't have to change it, it's just good to be aware of :)

**CSS**
- It's responsive and the accordion looks super - so no changes required!

jesinkey:
Hi! Updated according to the above and pushed to GIT :)

code/index.html
```diff
@@ -1,30 +1,117 @@
...
+<body>
+
+    <!--HERO-->
+    <div class='hero'>
+        <img class='heroimage' src='media/header.jpeg' alt='drag'>
+        <h1 class='logo'>Drag Doll</h1>
+        <ul class='menu'>
+            <li><a href='/Resources/'>Resources</a>
+                <li><a href='/Sign Up/'>Sign-up</a>
+                    <li><a href='/Q&A/'>Q&A</a>
```
HIPPIEKICK, line 24: These are siblings and should have the same level of indentation :)

HIPPIEKICK (approved):

PR #168, opened 2021-05-23
Week 3 - Business site - Malin Göthe

M4L5:


KarolCislowski (changes requested):
Your hero section is wider than the rest of your page what creates a white margin on the right side.
image

code/style.css
```diff
@@ -1,7 +1,201 @@
...
+.hero-section {
+  position: relative;
+  font-family: fantasy;
+  color: white;
+  margin: 0;
+  background: rgba(106, 142, 117, 0.7);
+}
+
+.hero-image {
+  width: 100vw;
```
KarolCislowski, line 14: do not use `vw` because the scroll bar makes your image wider than the visible part of the page, so it stretches the whole page on the right side.

KarolCislowski: I used `Pesticide` Chrome plugin and you can see that your hero is narrower than the image.
image

KarolCislowski (changes requested):
I think I found the issue.

code/style.css
M4L5: Hi Karol, thanks! I removed view width from the hero image (and the user section) and replaced it with %. I also downloaded the pesticide plugin. I've tried to recreate the white margin on the right side, using both vw and % but I still can't see it. I have also looked at the page from a pc laptop and can't recreate the margin. Best, Malin

KarolCislowski (approved):
Great Page 🎉

PR #174, opened 2021-09-03
Hedvig Mejstedt week 3

HedvigM:


code/index.html
```diff
@@ -3,28 +3,163 @@
...
+        </video>
+      </div>  
+    </header>
+
+    
+
+    <div class="middle-section">
+     <!-- Accordion -->
+     <section class="accordion">
+      <fieldset>
```
IdaAspen, line 40: Great that you have used the fieldset-tag to group the HTML form!

code/index.html
```diff
@@ -3,28 +3,163 @@
...
+
+            
+              <div class="radio-button">
+                <p>How would you like to spend the nights?</p>
+                  <div class="radio" >
+                    <input type="radio" id="tent" name="tent" value="tent">
+                    <label for="Something1">In a tent</label><br>
+
+                    <input type="radio" id="cabin" name="Something" value="cabin">
+                    <label for="Something2">In a cabin</label><br>
```
IdaAspen, line 138: I think some of the labels doesn't match with the ID they are pointing at. I'm not sure that is why I can choose to both stay in a tent and a cabin (or a hotel). Maybe something to look into?
I get following message"The for attribute of the label element must refer to a non-hidden form control."

IdaAspen (approved):
First, love that you dared to do the video instead of img and challenged yourself.  Also smart that you chose grid instead of Flexbox to be able to practice different methods. 
I think the result is super nice, and the video really sets the ”feeling” of the websites subject.  

The project meets all the requirements and works well on different screens, the form and accordion works!

The signup-form is really good with several input types and it feels really professional.
  I have already said it, but the "+" and "-" for the accordion is a really good feature and make it easier to understand that their is more to see if you click there, I will try that next time!
  Great that you have a well commented CSS file, it makes it so easy to read!  And the CSS file looks really good!! I have learned a bit about grid now from reading your code :)
  
  Good job!!

PR #187, opened 2021-09-04
Project-business-site-AnnaD

anndimi:


code/style.css
```diff
@@ -1,7 +1,286 @@
...
+  border: rgb(77, 106, 131);
+  text-align: left;
+  outline: none;
+  font-size: 18px;
+  transition: 0.4s;
+  border-radius: 10px;
+  font-family: 'Glory', sans-serif;
+}
+
+.question.one, .three {
```
Smelbows, line 164: Nice way to do the alternate colours. I think theres an 'even child selector' or something similar that you could use too. The benefit is that you could add 5 or 10 more questions and not have to write anything extra here to make them continue to alternate

code/style.css
```diff
@@ -1,7 +1,286 @@
...
+.active:after {
+  content: "\2212";
+  /* Unicode character for "minus" sign (-) */
+}
+
+.answer {
+  background-color: rgb(71, 131, 121);
+  max-height: 0;
+  overflow: hidden;
+  transition: 1s ease-out;
```
Smelbows, line 192: FYI Jacob and I were both having issues getting our slide out to work ,so we took a sneaky peak at yours to help us figure out what we were doing wrong :-)

code/style.css
```diff
@@ -1,7 +1,286 @@
...
+@media (min-width: 768px) {
+  body {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+  }
+  .slogan {
+    font-size: 20px;
+  }
+  .menu-btn {
```
Smelbows, line 226: I like your menu variation for mobile and desktop

Smelbows (approved):
Anna I think you've done a really great job with this week's project. Your CSS looks so smooth and professional. It seems as if you know exactly how to put it together so that it looks exactly right but hasn't used lots of 'forcing' to get it there. Specifically it looks like you've putt a lot of thought into how it will look on both desktop and mobile. You've added the styling required for a lot of the red/black requirements too. I put some specific comments on both the JS code and the form inputs. Consider adding some validation so that you get exactly the information that you're expecting in your HTTPBin. Well done! Sarah

PR #189, opened 2021-09-04
week 3: Business site Maria Petersson

hejmaria:
The project was to build a responsive site with a form, connected to httpbin.org, and build a accordion with JS. The header should include a big hero img or a video.

It was a lot of fun once I got started. I am new to javascript and I still have a hard time to understand how it works and how the tags are connected. I have used alot of stack overflow, w3 and random googling to wrap my head around it. 

I focused on handling one "problem" at the time. First the header, then the form and last the accordion. Due to learning by trial and error (and copy pasting) my code was a mess. the mobile first approach got a bit messed up. I need to be better at keeping it clean next week.

I had unnecessary problems with the accordion. If I had read the brief more closely I would have solved the problems quicker.

I used the picture tag for the header img, but switched in the end. It looked better with the background img, and was easier to style. I kept the picture tag in the code for my own remembering. 

I also have a problem with the width of the boxes in screen mode. Of course, I would want to fit it. The form-box isn't as wide as I would like. I've played around with margins and directions, but settled for the version that looked best. 

https://pensive-hugle-f55c00.netlify.app


code/style.css
```diff
@@ -2,6 +2,359 @@
...
+  display: inline-block;
+  font-size: 18px;
+
+}
+
+
+/* Accordion styling starts here */
+
+.wrap-accordion {
+  display: flex;
```
HedvigM, line 131: I don't really understand what you are flexing in "wrap-accordion" or what the intention is. You are flexing the body as well, and that's correct if you want to flex the accordion and the form.

code/style.css
```diff
@@ -2,6 +2,359 @@
...
+  margin: 30px auto;
+
+}
+
+.accordion {
+  background-color: rgb(236, 197, 241, 0.85);
+  border: 1px solid;
+  border-radius: 8px;
+  padding: 20px;
+  box-shadow: 5px 5px 15px 5px rgba(0,0,0,0.5);
```
HedvigM, line 142: Really nice effect!

code/style.css
```diff
@@ -2,6 +2,359 @@
...
+
+footer{
+  background-color: rgba(0, 0, 0, 0.3);
+  position: relative;
+  width: 100%;
+
+}
+
+.grid-container {
+  display: grid;
```
HedvigM, line 203: Nice! You used both grid and flex in this project! =)

HedvigM (commented):
Really Nice job! 
Your web page looks so nice, I get so inspired to keep work on mine to make some of the features you have incorporated. 

I have learned a fair amount by reading your code, so thank you! The first thing I had to check out when I saw your page was how you managed to make the plus and minus sign to stay at the right at all times. 

Great to see how you also have the email validation!

Maybe I've should have pointed at the document as well, but you hade a border with 1px on the .accordion, without any color attached to the border, that means that the border isn't showing at the moment. 

Really great work! You have managed to make an incredible nice site!

PR #195, opened 2021-09-05
Kara Howes Business site (project 3)

KaraHowes:
I wanted to create a business for selling plants. I thought this would be easy to obtain nice free images and should ne workable. The site should have a sign-up form, a Hero image and an accordion styled with JS.

I approached the problem using a mobile-first methodology and flex box. The idea was for the content of the accordion and sign-up form to not vary, only their positions on the website. I used JS to make the accordion interactive AND also the form interactive as well. I custom styled the sign-up checkboxes and radio buttons. I did want to add an icon to the accordion, but I didn't have time.

https://musing-elion-dfefd6.netlify.app


code/style.css
```diff
@@ -1,7 +1,503 @@
...
-in this case it is the div with the answer class next to the div with the active class.
-this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
+
+
+* {
+  margin: 0;
+  padding: 0;
+  
+}
+html {
```
nama0027, line 8: i think in one of lectures Jennie suggests that instead of styling <html> use <body> as html tag refers to whole html including head

code/style.css
```diff
@@ -1,7 +1,503 @@
...
+  height: 100vh;
+  font-size: 18px;
+}
+
+/*** Body ***/
+body {
+  height: 100%;
+  box-sizing: border-box;
```
nama0027, line 16: I am not sure but generally its better to use universal selector for box-sizing-

nama0027 (commented):
Good job! your code is working perfectly and you have fulfilled all requirement. and the custom radio and check buttons are so nice. I really like the description you provided while submitting the pull request. It is through and clear!

In order to improve the code readability and understandability, I would suggest that you should try to add more comments. In addition to this may be remove the extra empty lines would also make the code look better overall.
Another thing what I would like to suggest is about using class name for styling. I think its consider a good practice to use class name while styling instead of tag name.

PR #207, opened 2021-09-05
Sarah Mottram Coffee: Project-business-site

Smelbows:
I was really pleased with the amount of new CSS I learnt this week and how much I was able to apply to this project. I think I've hit all the blue requirements and some of the red/black also. I was really pleased with the styling of the accordion and the slide down effect, and also the different form inputs that I tried. I would have liked to do more with the JS this week but ran out of time. The main down-side of this weeks code is the lack of comments. I ran out of time to add them at the end, and from next week I plan to add comments as I go along which I think will help me as I'm coding, as well as help anyone who's reading the code!! I think I can also work on how I name classes, so that my CSS doesn't have to be repetitive. I have some ideas of how to do this ready for next week.


code/style.css
```diff
@@ -2,6 +2,353 @@
...
+
+.fieldset {
+  border-radius: 10px;
+  display: flex;
+  flex-wrap: wrap;
+  margin-bottom: 10px;
+  
+}
+
+.input, .select-input {
```
madeleinesvensson, line 140: Nice use of one style with double classes :)

code/style.css
```diff
@@ -2,6 +2,353 @@
...
+  display: flex;
+  justify-content: space-between;
+  font-weight: bold;
+  margin-top: 30px;
+  padding: 0 30px;
+  background-color: rgb(244, 245, 238);
+  border: black solid 1px;
+}
+
+.question:hover, #submit-button:hover {
```
madeleinesvensson, line 221: I LOVE THE COFFEE POT AS A CURSOR <3

code/style.css
```diff
@@ -2,6 +2,353 @@
...
+  color: white;
+  background-color: rgb(66, 39, 7);
+  cursor: url(./pictures/icons8-french-press-48.png), auto;
+}
+
+.question:nth-child(4n+4) {
```
madeleinesvensson, line 227: :nth-child (4n+4)??? what does that mean? could maybe use a comment just to explain what this does

code/style.css
```diff
@@ -2,6 +2,353 @@
...
+  transition: 1.3s ease-in;
+}
+
 .active + .answer {
-  /* some styling here */
+  max-height:200px;
+  opacity: 1;
 }
+
+/* .right-aside {
```
madeleinesvensson, line 266: I guess these could be deleted since they're commented out?

madeleinesvensson (commented):
So Sarah, very good job! You said that the code is messy under the cover but I don't think it's messy! The structure is very good and names makes sense so I understood most of it (there were a few things I found strange(that is commented) but I might have missed where the class is etc) Other than a few places the code is easy to follow along and written in a way that is tidy! I mean the label and input structure is amazing! I thought the css was easy to follow because it sort of went in descending order(on purpose or not, very good) I could not find anything broken and everything works as far as I can tell! As you wrote I think this definitely passes general and blue requirements, and some of the higher ones as well :P! Good job!

PR #219, opened 2021-09-05
Knit happens - Nina Rivera

NinaAlejandra:


code/script.js
```diff
@@ -1,9 +1,15 @@
...
 
-// Selects an HTML element, and calls a function which will be executed when the element is clicked.
-document.getElementById("section1").onclick = toggle
+// Still kind of confused about how this works
+document.getElementById('section1').onclick = toggle
+document.getElementById('section2').onclick = toggle
+document.getElementById('section3').onclick = toggle
+document.getElementById('section4').onclick = toggle
+document.getElementById('section5').onclick = toggle
+document.getElementById('order').onclick = toggle
```
Groovedharry, line 12: This is totally different from what I did but I understand how it works and its interesting seeing you doing something totally different from me but gaining the same result!

code/style.css
```diff
@@ -1,7 +1,188 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: cover;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: relative;
+  overflow-x: hidden;
+  flex-direction: column;
+}
```
Groovedharry, line 45: What does "position: relative;" do here? tried to uncheck it in inspect but it doesnt really do anything, dont think 
overflow-x :hidden; does anything either I have seen some position: relative in your code but dont really see what they do, will mark them up so you can see where they are

code/style.css
```diff
@@ -1,7 +1,188 @@
...
+
+.parent {
+  display: flex;
+  text-align: center;
+  width: 80vw;
+  margin: auto;
+  margin-top: 5%;
+  flex-direction: column;
+  justify-content: center;
+  position: relative;
```
Groovedharry, line 61: Position relative, what does it do?

code/style.css
```diff
@@ -1,7 +1,188 @@
...
+  display: flex;
+  flex-direction: column;
+  margin: auto;
+  margin-bottom: 10%;
+  padding: 5%;
+  width: 75vw;
+  font-size: 12px;
+  background-color:darkseagreen;
+  opacity: 1;
+  position: relative;
```
Groovedharry, line 86: what does it do?

code/style.css
```diff
@@ -1,7 +1,188 @@
...
```
Groovedharry, line 111: what does it do?

Groovedharry (approved):
Your code fulfill the basic requirements for the assignment and the code is easy to read and follow.
I learnt a few things from you code that I didn't think of doing, Great job!

One thing you could've worked on is in the form to align the items and make it more centered but I believe you said in your readme that you would've done it if you had more time

PR #225, opened 2021-09-06
Kriss

silvertejp89:
Struggled wirh time this week, but rally produkter that I got sametinget, and that I asked for help in my group!


efstasia (commented):
Overall a really good job even though you said you've been struggling with time management! Just some minor things that you could fix, but nothing mayor. The project was responsive on all platforms. You did all the blue requirements and i think that you should be proud!

PR #226, opened 2021-09-06
Business site - first try

pcruzem:
The project is far from finished, still haven't figured out how everything works.


Maja473 (changes requested):
The website looks really cool with the hero video, and the sign up form is really well done! It checks that the e-mail address must be functional and upon clicking 'register' sends everything to the right place. The inclusion of choosing an age is neat, and I think you've done a great job overall. However, I can't click the accordion, but I assume that's because as stated, everything isn't quite finished. I'd suggest making sure the accordion works, since one of the requirements is building it with Javascript, and that seems to be the part that isn't currently functional if I've understood it correctly. This project was very difficult and you've managed very well, but that'd be my main suggestion. Good job!

PR #230, opened 2021-09-06
idaUtterström-business-site

Idautterstrom:


code/script.js
```diff
@@ -1,9 +1,10 @@
-// A function that adds and remove the class "active" on the section you click on.
-// We haven't really talked about ´this´ yet, but we will... ;)
-// console.log(this) to see how it works
 function toggle() {
-  this.classList.toggle("active")
+  this.classList.toggle("question")
```
gustavfrid, line 2: Nice workaround in combo with line 74.

code/style.css
```diff
@@ -1,7 +1,166 @@
...
+.answer {
+  display: none;
+  color: rgb(135, 133, 133);
+  padding: 0px 40px 15px 40px;
+  text-align: left;
+  margin-left: 40px;
+}
+
+.question + .answer {
+  display: block;
```
gustavfrid, line 75: I think this overwrites the styling above and keeps the .answer class visible. If this is what you intend I think you can write display: block; in the class above instead of display: none; and just remove these lines.

However if you intend to have the .answer hidden from the start you need to use some other selector here. Maybe a .active + .answer selector and then toggle .active in the javascript?

code/script.js
```diff
@@ -1,9 +1,10 @@
-// A function that adds and remove the class "active" on the section you click on.
-// We haven't really talked about ´this´ yet, but we will... ;)
-// console.log(this) to see how it works
 function toggle() {
-  this.classList.toggle("active")
```
gustavfrid, line 5: You can use this one if you want to use the  .active + .answer selector to toggle the accordions visibility to hidden in the start se comment on lie 74

gustavfrid (approved):
Overall a very nice business site, I really like your styling. Looks very professional and I get the impression that you obviously have a talent for the styling.

The only thing that confused me was that the accordion was visible from the star but you have a styling that says that it should be hidden. I hope my comments was not to confusing.

PR #232, opened 2021-09-06
Project Business Site - Jessica Nordahl

JessicaNordahl:
This project was to create a business-site

It contains an accordion and a form

I used JavaScript, CSS and html, 
Background-image comes from unsplash


JessicaNordahl:
I've managed to do some small changes. To be continued :)

Alisebrink:
I've given your project a look and first of all, good job! 
I know that it is not finished yet, but I will give you a few comments still. 

I can see that you have styled your page, you have a signup form and an FAQ accordion that works. Good job!

A few pointers though. I think that it is easier to make a hero image if you use the background-image attribute, instead of putting an img-tag in the hero-box. 
It looks like this:
`.your-container
background-image: url(/yoururl.jpg);
`
This way you can style your container and make it fit the hero-box. Please ask me if you have any questions. 

I also can see that your page is not yet responsive, but I believe that it may be a feature that your are working on?

code/index.html
```diff
@@ -4,27 +4,88 @@
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Project 3</title>
+    <link rel="stylesheet" href="style.css">
     <!-- dont forget to add a css file and link it here! -->
   </head>
   <body>
-    <!-- video or image as a header is cool :) -->
+      <header>
```
Alisebrink, line 11: This is where you would put the hero image. Instead of doing a img tag within the header tag you can use the background-image attribute. Using that wont make your image stretch like it does now when you expand the window.

code/style.css
```diff
@@ -2,6 +2,139 @@
...
+
+header {
+  display:flex;
+  width: 100%;
+  height: auto;
+  text-align: center;
+  background-color:rgba(40, 233, 207, 0.753);
+  color:white;
+  position:fixed;
+}
```
Alisebrink, line 17: This is where you would style your background image. Put it as `background-image: url(yourimage.jpg);`

You can also make it fit to size and scale with the browser. Here you can read some more on it:
https://www.w3schools.com/cssref/pr_background-image.asp

PR #234, opened 2021-09-06
Business-site-Aleksandra

A1eksa:


silvertejp89 (commented):
God job! The project seems to meet all requirements. Little tricky with the responsive design but that´s it! I thought I could read your code and must say I find the rounded boxes really decorative.
 I couldn´t see a lot of pushes, that is something I think one could do more often as a good habit. I struggled with this task myself so I´m impressed woth your result! Thank you for sharing!

PR #243, opened 2022-01-26
Finished the proyect

Dopest0727:


nadialefebvre (commented):
You website (and your code) are both great, Mauri! It's styled, there is a responsive video in the header, a signup form with more than 3 input types and a FAQ accordion. You went over the minimum requirements, for example with the icon changing when the accordion section expands/collapses and the mouse cursor changed. You can see that I only had tiny thing to say about your code, like kebab-case for classes. Bravo!

PR #251, opened 2022-01-28
Coffee-business-site

savannah-hayes:


Nosslexa (commented):
I followed these questions when looking at your code: 
​​1. Does it work or does anything look broken?
2. Am I able to understand the code easily?
3. Does the code have lots of duplicates? Could it be tidied up?
4. Does it follow the general and blue requirements?

1. The code works perfectly and so does the responsiveness. 
2. Your code looks very tidy and is easy to follow.. Could not find any duplicates.  
3. your code is easy to understand and thanks to your structured way of naming and giving id’s i can almost predict for example, what elements are going to get styled together. 
If I understand your code correctly I see that you use js to add validation to the sign up where the email must have a @ and that the phone number only accepts numbers 0-9. -very nice!
4. you have met all the requirements- great job!

Great job! I think your site looks great and so does your code. Your structure and naming makes it really easy for me to understand what's going on even though it's more advanced than my own knowledge and that means your code is easy to follow for anyone who would potentially take over after you. later on i hope i can give you a more constructive review but for this time, very well done! I will try and take after your example in my own coding.

PR #253, opened 2022-01-28
project-business-site/frida 

Nosslexa:


code/style.css
```diff
@@ -1,7 +1,232 @@
...
+}
+
+h2{
+  margin: 0.5rem;
+  color: #a96e52;
+}
+
+/* signup form and accordion section*/
+
+@media (max-width: 767px;){
```
jenniferwvng, line 60: ```@media (max-width: 767px;){``` Seems like a semicolon accidentally got in here :)

code/style.css
```diff
@@ -1,7 +1,232 @@
+*{
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+  -webkit-font-smoothing: antialiased;
+  text-rendering: optimizeLegibility;
```
jenniferwvng, line 6: ``` -webkit-font-smoothing: antialiased; text-rendering: optimizeLegibility;```Cool! I never heard of these

code/style.css
```diff
@@ -1,7 +1,232 @@
...
+  font-size: 12px;
+}
+  
+
+/* section for styling of accordion */
 /* the + sign is used to select the sibling that comes immediately next to the element you selected.
 in this case it is the div with the answer class next to the div with the active class.
 this can be used for the FAQ-accordion. read more here: https://www.w3schools.com/css/css_combinators.asp */
 
 .active + .answer {
```
jenniferwvng, line 166: ```.active + .answer { display: block;  }``` In the desktop version of the page, the accordion content stretches outside the content edges of its container. Try to add overflow: hidden; and see if that solves the problem.

code/index.html
```diff
@@ -4,27 +4,98 @@
...
-      <div class="question" id="section1">Section 1</div>
+
+
+<!-- Signup form -->
+<section class="content-wrapper">    
+  <Div class="form-wrapper">
+    <form action="https://httpbin.org/anything" method="post">
+      <h1>Sign up now</h1> 
+        <label>
+          <input type="text" name="name" id="name" placeholder="Name">
```
jenniferwvng, line 42: Could try to add 'required' tag in input to prevent the form from submitting empty of content. The form works fine as it is now but this could improve user interaction :)

code/index.html
```diff
@@ -4,27 +4,98 @@
...
+          Coffe is life
+        </label><br>
+        
+        <label>
+          <input type="radio" name="radio" id="ocupation">
+          Not without my Tea
+        </label><br>
+        
+      <br><label class="accept">
+          <input type="checkbox" name="accept" id="checkbox">
```
jenniferwvng, line 64: Its generally considered to avoid using <br> if there are other alternatives, and if used it's when there is a need to make a line break inside a text. For styling and creating spaces between sections/in this case labels, it's better to use the margin CSS property. So in this case I think we could solve this with applying CSS styles instead.

jenniferwvng (commented):
Summary: 
Cute page content, form is working as it should, could add some ```required``` tags to improve user experience (avoid submitting empty fields). Great work with making the hero, sign-up section and FAQ section mobile responsive!

The FAQ content stretches outside its container, but should be easily fixed with the suggestion of using ```overflow: hidden```. 

Well done!

code/style.css
Nosslexa: Added oveflow: hidden; but still have the same problem.  Will se if i can find another solution.

PR #254, opened 2022-01-28
Business site - week 3

SofiePellegrini:


code/style.css
```diff
@@ -1,7 +1,327 @@
...
+}
+
+input[type=radio] {
+  
+  display: inline-block;
+  margin:0px 0px 25px 0px;
+}
+/*Accordion*/
+
+.accordion{
```
Laura-Sjolander, line 124: The accordion container expands both up and down when you open the accordion, it would make more sense if it only expand downwards, maybe something to do with overflow?

Laura-Sjolander (commented):
**Nice job Sofie!** In general a very pleasant website for flowers. Nice colors, nice design and cudos to you for making the video work so well. 👍 

Good thing you have comments so I can se what is what in the code and that you had some information on the project in the the Read me- file. 

Only thing I note, is that when the accordion opens up , the whole container grows with it both upwards and downwards.  The plus sign that opens up the accordion is also very small so you don´t notice it at first (or is it me, needing stronger reading glasses 😃 )

There is some extra pictures in the images folder that I see here in the github file that I think you are not using that I guess you could delete from the folder before pushing it up to github.  

Nice and sublte animations - hover effect- on the buttons for newsletter and the icons in the footer. 

You surely meet the requirements on this weeks assignment and I really like this website, 
Keep up the good work. 👏

PR #257, opened 2022-01-29
Week 3 business site

SteppbySteph:
Here is my version of week three repo:  https://frusto.netlify.app


code/index.html
```diff
@@ -3,28 +3,131 @@
...
+            </li>
+          
+          </ul>
+          </nav>
+      </div>
+
+      
+      <div class="logo-place">
+        <a href="#" class="logo-mobil">Frusto</a>
+      </div>
```
michaelchangdk, line 45: I was wondering why you had your logo twice - once as `.logo` and once as `.logo-mobil`. I saw that you used `display:none` on the `.navbar`, which hid your logo as well. One potential solution in the future would be to use `display:none` on your `.nav-item` class instead, which would hide the links, so you don't have to insert the logo twice in your HTML and CSS. Another option is to take place the logo before the `<nav>` class, and then hide the entire `.navbar`. That way, you don't have to be worried about the logo being duplicated twice in the code, and any styling can be solved using @media queries.

code/index.html
```diff
@@ -3,28 +3,131 @@
...
+        
+          <label for="password">Password</label>
+          <input id="password" type="password" name="password" required/>
+        </div>
+      
+   
+        <div class="radiobutton">
+          <div>
+            <input id="prof" type="radio" name="occupation" value="prof"/>
+            <label for="prof">young and restless</label>
```
michaelchangdk, line 74: The `value="prof"/"student"` and `name="prof"/"student"` combined with the `label="young and restless"/"tired and frustrated"`, also on lines 79 & 80, make it a bit confusing when sending in the form to see what's actually been chosen. Maybe a slightly better choice of "name" and "value" could have been chosen that fit the label better?

code/index.html
```diff
@@ -3,28 +3,131 @@
...
-    <section class="accordion">
-      <!-- Accordion -->
+    <!-- Accordion -->
+    <section class="accordion-container">
+    
+      <p>So nice that you have made it here! This is an online community for people who feel they are tired of everything around them including themselves. 
+        The goal with this website is to gather people who want to make a difference for oneself and others.
+        Put simply, this is a place for frustrated people to unload their feelings. </p>
+      <br>
       <h1>FAQ</h1>
```
michaelchangdk, line 104: `<br>` is maybe not the best choice, convention wise, for wanting to put some space between your `<p>` tag and your FAQ header. I think I would maybe have added a class to your FAQ header, and added either some margin or padding to the top and bottom to create a bit of space around the FAQ.
By adding a class to the `h1` tag for the FAQ, you could also `text-align: center` it, as well as change the font size to be more similar to the accordion headers.
You could also have added a class to your `<p>` tag, which would have allowed you to add the line-height property, which makes it a bit more legible / adds some space between the lines of text.

code/index.html
```diff
@@ -3,28 +3,131 @@
...
+  
+      </form>
+    </section> 
+
 
-    <section class="accordion">
-      <!-- Accordion -->
+    <!-- Accordion -->
+    <section class="accordion-container">
+
```
michaelchangdk, line 99: I was missing a bit of a header here for the introduction to your business - just a personal choice and not a critique of your code in any way.

code/index.html
```diff
@@ -3,28 +3,131 @@
...
+        <p> The membership is free but we do take a small charge for private sessions.  
+        </p>
       </div>
+  
     </section>
-    <!-- dont forget to link the javascript file here! -->
+  </main>
+
+    
+  <!-- dont forget to link the javascript file here! -->
```
michaelchangdk, line 127: This comment could be deleted or changed to `<!-- Link to JavaScript file here -->`

code/index.html
```diff
@@ -3,28 +3,131 @@
...
+        </p>
       </div>
 
-      <div class="question" id="section2">Section 2</div>
+      <div class="question" id="section3">Does it cost to be a member?</div>
       <div class="answer">
-        <p>Lorem ipsum...</p>
+        <p> The membership is free but we do take a small charge for private sessions.  
+        </p>
       </div>
```
michaelchangdk, line 121: Super clear and easy to see what's going on in the accordion. Great job!

code/script.js
```diff
@@ -1,9 +1,16 @@
...
 function toggle() {
-  this.classList.toggle("active")
+   this.classList.toggle("active")
+  //  console.log(this) Visar i inspekt konsollen vad som händer när funktionen åberopas
 }
 
 // Selects an HTML element, and calls a function which will be executed when the element is clicked.
 document.getElementById("section1").onclick = toggle
+document.getElementById("section2").onclick = toggle
+document.getElementById("section3").onclick = toggle
```
michaelchangdk, line 13: Simple and effective. Love it!

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+
+  .logo-mobil {
+    display: none;
+  }
+
+  .navbar { 
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+    padding: 20px;
```
michaelchangdk, line 222: I would have added the padding here on `.navbar` to the `.inner-hero-header` element instead - I can see that your logo and links are vertically aligned in the center when I use Pesticide, but when the naked eye perceives it, it looks almost as if it's just a bit off center (vertically). I think that might have something to do with the padding extending past the actual `.inner-hero-header` borders. 
This mostly only becomes a "problem", or breaks the navbar a bit, when you make your browser very short. The `.inner-hero-header` has a height of 8vh, and when the padding is in it's child, that means the inner-hero-header becomes shorter than the text overlaying it, and it ends up laying halfway over the gray opaque background, and halfway over the hero image.

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+    margin-top: 0;
+  }
+    
+  .logo-place {
+    display: none;
+  }
+
+  .logo-mobil {
+    display: none;
+  }
```
michaelchangdk, line 216: Small comment - I don't think you need to add `display: none;` to `.logo-mobil` when you already have it for it's parent div, `.logo-place`. So this could probably be deleted without changing your site.

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+  justify-content: space-evenly;
+}
+
+.inner-hero-header {
+  width: 100vw;
+  height: 7vh;
+  background-color: gray;
+  color: #fff;
+  opacity: .7;
+  margin-top: 0;
```
michaelchangdk, line 35: LOVE that you used opacity, as well as vh units.

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+  flex-direction: column;
+  justify-content: top;
+  height: fit-content;
+  border-radius: 15px;
+  background-color: #E2E5DE;
+  padding: 25px;
+  margin-top: 10px;
+  margin-left: 5px;
+  margin-right: 5px;
+  margin-bottom: 10px;
```
michaelchangdk, line 74: I would have added a teensy bit more margin for the mobile version - it looks a little squished up against the sides of the viewport. This is just a personal preference though :)
You can also use shorthand for the margins here like this:
`margin: 10px 5px;`
When you specify only two values, the first one is top and bottom margin, and the second value is left and right values.

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+  padding-left: 20px;
+  padding-right: 20px;
```
michaelchangdk, line 76: In the same way that you can use shorthand for the margins in my above comment, you could do the same for padding here. So this could be achieved by writing:
`padding: 0 20px;`
Since you want 0 padding on the top, but 20 px on the left and right.

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+}
+
+.form-wrapper h2 {
+  text-transform: uppercase;
```
michaelchangdk, line 80: Super cool - I was reading your labels and wondering - how are they uppercase on the site but lowercase in your HTML? I didn't even know this was possible! I'm definitely stealing this in the future.

code/style.css
```diff
@@ -2,6 +2,340 @@
...
+  font-size: 1rem;
+}
+
+.form-inputs {
+  display: flex;
+  flex-direction: column;
+  margin-top: 5px;
+  font-size: 1.5vh;
```
michaelchangdk, line 88: Personal preference again - but the font sizes for the sign-up form are a bit too small, especially compared to the accordion and introduction. On mobile they look a little bit squished and hard to read.

code/index.html
```diff
@@ -3,28 +3,131 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="style.css">
+    <link href="https://fonts.googleapis.com/css2?family=Montserrat&display=swap" rel="stylesheet">
```
michaelchangdk, line 7: I would maybe have added comments here `<!-- Link to Stylesheet -->` and `<!-- Link to Google Fonts -->` to make it easier to find for your future self.

michaelchangdk (commented):
⭐ ⭐ ⭐ ⭐ ⭐  5 Stars!!! ❤️ Michael

PR #261, opened 2022-01-29
Farmers Academy - project business site 

Josse79:
A fun project but also challenging!
Struggle with CSS and styling.
Problems with video autoplay after publishing.
But learnt alot overall:-)


amandatange (commented):
I think you've done a really good job with making a responsive business site with not only a hero video but a form, FAQ accordion with js and also styled everything very nicely! 

What I would suggest for the future is adding a couple of max/min values to eg font-sizes and heights etc to keep it responsive whilst making sure that nothing gets too small or too big. 

The website works well and the code is not commented a whole lot but it is very easy to read and understand nevertheless! It absolutely fulfills the requirements!

PR #268, opened 2022-01-29
Project Business Site Week 3

joannalodell19:


code/index.html
```diff
@@ -2,29 +2,117 @@
...
+              <h2>FAQ</h2>
+
+              <div id="questionBox">  
+                <div class="question" id="question1">Why should I hire Pepped?</div>
+                <div class="answer" id="answer">
+                <p>First of all, by the sound of our name - we guarantee that you will get more pepp in your life.</p>
+                </div>
+              </div>
+
+              <div id="questionBox">
```
Katarina821, line 89: Maybe it would be better to use class instead if Id for `#questionbox`? I think you are only supposed to use id when its one specifik element.

code/style.css
```diff
@@ -1,7 +1,428 @@
...
+  margin-top: 5%;
+  margin-bottom: 5%;
+}
+
+/* ACCORDION CONTAINER */
+.accordionWrapper {
+  margin-top: 0%;
+  /* margin-left: 3%; */
+  width: 70%;
+  flex: 1;
```
Katarina821, line 376: Had to google the` flex:1` and learned something new, thanks:)!

Katarina821 (approved):
Hi Joanna!
Great job with this weeks project, the code is easy to read and well structured, the deployed page looks good and the responsivness works fine. Also helped alot with the comments, made the code easier to follow. Added a few comments, hope they can help. Well done!

PR #269, opened 2022-01-29
Welcome to Hanna's VIP event

HWallberg:
Link:
https://thirsty-carson-416bac.netlify.app/

Struggled the most with the containers and divs, but I think I got it right in the end. I hope it's understandable.


eyahya-khan (commented):
Need some line height/space between line

PR #272, opened 2022-01-29
added html css and js

LovisaBrorson:


code/style.css
```diff
@@ -1,7 +1,211 @@
...
+
+body {
+  font-family: 'Source Sans Pro', sans-serif;
+  background-color: rgb(243, 215, 178);
+}
+
+
+.grid-form {
+  display: flex;
+  flex-flow: column;
```
kolkri, line 20: I was wondering why you used flex-flow here, should it be flex-direction instead? It works anyway but with flex-flow you can define flex-direction and flex-wrap at the same time, for example flex-flow: column wrap;

code/index.html
```diff
@@ -4,27 +4,83 @@
...
+      </label>
+      <label> 
+        <br>
+        <input type="tel" name="telefonnummer" placeholder=" Telefonnummer" required />
+      </label>
+      <label class="labelmargin"> Något vi bör veta? 
+        <br>
+        <textarea></textarea>
+      </label>
+      <label class="labelmargin"> <a href="/link/"> Jag har läst och godkänner vilkoren</a>
```
kolkri, line 50: if you are just making fake link and it does not actually lead anywhere you can use for example sign # like this:
<a href="#" id="theLink">My Link</a>

code/style.css
```diff
@@ -1,7 +1,211 @@
...
+.accordion::after {
+  content: '\25be';
+  float: right;
+  transform: scale(1.5);
+}
+
+.accordion--activ {
+  background-color: gray;
+}
+
```
kolkri, line 150: You have typing error here, it should be .accordion--active, you are calling this style in javascript function (to make the arrow pointing up when accordion is open), but it does not happen cause this typing error. You could fix that and see if it's working then :)

kolkri (commented):
The header with the video is really nice! Your code is clean and easy to read and find things, You used comments in css file which made the code easy to follow, you could use comments even more. I added some small comments that came to my mind, and I found one typing error which effects to your accordion when you open it (look below the comments). Really great job, if you have time later you can try some form styling for example. The page has the minimun requirements of the week 3, which were:
- Style your page
- Site should have a header with responsive image/video
- Site should have a signup form with at least three different input types
- Site should have an FAQ accordion, built with JavaScript. 

One thing you are missing, is that you could fill the READ ME -file. It is very useful for people who later look at your github so they know what this project is about and it also helps you when going through your project later! :)

LovisaBrorson:
Hi,

I took the grid from this page we got from Karin,
An Example of Nesting Flexbox inside CSS Grid (codepen.io)
<https://codepen.io/JennieDalgren/pen/LYLEjQK?editors=0100>
 I was just very happy it worked and did not reflect over why they
choosed flex-flow.
I do not know, flex that well yet, so I don't even understand this "It
works anyway but with flex-flow you can define flex-direction and flex-wrap
at the same time, for example flex-flow: column wrap;"

I hope it will come to me... but I will change it to flex-direction.

Den tis 1 feb. 2022 kl 21:27 skrev kolkri ***@***.***>:

PR #273, opened 2022-01-29
Project week 3 - Business site 

jenniferwvng:
Link to my deployed project: https://confident-mayer-dad98b.netlify.app/

My reflections on how this week's project turned out:
Great, feels like I learned a lot new about CSS selectors and fun to include JS.

Things I'd like to have clarified or explained in more detail:
None of which I can think of rn


bostrombundock:
This project is very well done, since I am totally beginner to programming this might not be a code rewiev with lots of input of what could be better or different, but I must say the codes are very clean and it´s easy even for a beginner to follow and understand what is actually happening in the code, and everthing in Html, CSS and JS The Js are so clear and connected in a way without too much codes in for example Css. I´m very impressed and it is easy to use, accordion is really nicely done and the whole website catches the eye direclty!

jenniferwvng:
> This project is very well done, since I am totally beginner to programming this might not be a code rewiev with lots of input of what could be better or different, but I must say the codes are very clean and it´s easy even for a beginner to follow and understand what is actually happening in the code, and everthing in Html, CSS and JS The Js are so clear and connected in a way without too much codes in for example Css. I´m very impressed and it is easy to use, accordion is really nicely done and the whole website catches the eye direclty!

Thanks a lot! :)

PR #279, opened 2022-01-30
Week 3 Business site Amanda Tange

amandatange:


mistscale (commented):
Good job with this project Amanda! Your site is beautiful as well as your code. The code was easy to follow and very good use of comments in the CSS. Responsiveness works well and I really like that you used color picker (I guess) on the header to pick matching colors for the accordion, form and footer.

PR #284, opened 2022-01-30
project-business-site

kolkri:


tiiliu (commented):
Code review:

The indentation of your code is good and clear. It's easy to recognize the different sections inside your code. I didn't see really a problem with this in your code, but if you would have a lot of different attributes for one element, one option to even further improve the readability of the code is to place the attributes of the element on different rows, for example like this:

img 
	class=””
	src=””
	alt=”” />

or 

input
	required
	id=””
	class=””
	value=””
	name=””
	type=””

I noticed that you had used camelCase in all the naming of your code. In naming html and css classes and ids you should aim to use only kebab-case and use camelCase only when naming Javascript variables and functions.

Your way of having the media queries under the elements of which they were referring to was interesting! I've placed them separately in the end of the css file but seeing a different way got me contemplating what would be the best solution for their placement. Because I can imagine there's benefits in having them next to the specific element too, maybe especially when you need to go back and change something.

You've met the blue level requirements with your project and also achieved advanced goals too. 💯  All in all, your site looks beautiful and the functionality of it is very good, everything worked flawlessly. Also the responsiveness is there, the site works well and looks good in mobile and desktop. The message after submitting the form is very cool. Great job! 👍  😊

PR #286, opened 2022-01-30
Business Site

JenFi72:
Project where we were suppose to build a business site including:
- Image or video header
- Signup form 
- Accordion

I chose to build a site where you buy flowers. 

**Problems**
I tried to use the "Mobile First"-approach when starting the projects. My main goal was to add the JS code for accordion in a good way, though I struggled a bit. Although I managed to solve the problems by thoroughly follow different instructions films from Youtube. If I would do the project again I would have started with the form earlier - I struggled a bit to style it in a user friendly way. 

View it live: https://heuristic-heisenberg-940b6f.netlify.app/


style.css
```diff
@@ -0,0 +1,258 @@
...
+    }
+    .answer{
+      font-size: 18pt;
+    }
+    }
+  
+  
+    @media (min-width: 1024px) {
+  main{
+  flex-flow: row-reverse;
```
jiiinCho, line 201: Wow! so carefully designed in mobile screen, amazing!

index.html
```diff
@@ -0,0 +1,186 @@
...
+ 
+  <div class="intro">
+    <text>Do you love flowers but sometime do not have time to buy them? Flower Power Inc. brings you the nicest flowers in season picked by our star florists. The bouquets are delivered in a box outside your door. Read below to find out more. <strong>ENJOY!</strong> </text>
+  </div>
+
+  <main>
+
+ <!--Accordion-->
+
+  <div class="accordion">
```
jiiinCho, line 37: maybe try semantic elements like <section>, <article> instead of <div> as the technigo lectures suggested? I heard it helps web browser to understand html better :)

jiiinCho (commented):
A nicely design webpage with diverse fonts and colours, great work! Your website works well all screen sizes without any issues. I think you achieved all blue level requirements and red level requirements (except having different background color every other rows). Below is in details: 
* Style your page <completed>
* Site should have a header with responsive image/video <completed>
* Site should have a signup form with at least three different input types <completed: text, email, radio, checkbox>
* Site should have an FAQ accordion, built with JavaScript <completed : accordionItemHeaders >
* Try out some other form elements or input types. <completed : label, button>
* Add an icon before the title in the accordion indicates whether the section is expanded or collapsed.  <completed : accordion-item-header.active in css>
* Add some validation to the form <completed: type attribute validation in input element> 
* On hover, change the mouse cursor : <completed : .accordion-item-header>

One minor comment would be trying 'required' attributes one of your input element. Because current form can send an empty form to  https://httpbins.org/anything, which is, of course, not a problem at all!  But it might be good to try at this point as the lecture about form validation w3 suggested :)

PR #291, opened 2022-01-30
Sofia Ringstedt project week 3

sofiaringstedt:
https://friendly-mclean-0dc084.netlify.app/


code/script.js
```diff
@@ -1,9 +1,8 @@
...
-}
+const accordion = document.querySelectorAll('#button') 
 
-// Selects an HTML element, and calls a function which will be executed when the element is clicked.
-document.getElementById("section1").onclick = toggle
+accordion.forEach((accordion) => {
+  accordion.addEventListener('click', () => {
+    accordion.classList.toggle('open')
+  })
+ })
```
ariallahyar, line 7: In order to not get confused with what is what, try this:
- keep your function named `accordion` 
- rename the parameter i.e. what is within the function's parenthesis `(accordion)` to something more descriptive like `accordionButton`

Example below. Remember that you can write literally anything as the parameter to describe what it is. So what you're code is saying is in the accordion, for each accordion button, do something when the button is clicked.
 ```javascript
accordion.forEach((accordionButton) => {
  accordionButton.addEventListener('click', () => {
    accordionButton.classList.toggle('open')
  })
 })

code/style.css
```diff
@@ -1,7 +1,194 @@
...
+  list-style: none;
+  color: white; 
+  margin: 1rem;;
+  font-weight: bolder;
+}
+
+.title {
+  position: absolute; 
+  display: flex;
+  flex-direction: column;
```
ariallahyar, line 70: These three lines can be removed from `.title`:
```css 
position: absolute; 
display: flex;
flex-direction: column;
```

`justify-content: center;` is already applied to the `.image-header` div.  Your title will be centered horizontally since it is within that div. Then, in order to center it vertically, you will want `align-self: center;`

code/style.css
```diff
@@ -1,7 +1,194 @@
...
+  justify-content: center;
+  position: relative;
+  height: 55vw;
+  width: 100vw;
+  background-image: url('hero.png');
+  background-size: cover;
+  background-repeat: no-repeat;
+  background-position: center;
+  overflow: hidden;  
+  text-align: center;
```
ariallahyar, line 27: There is no text in the `.header` so `text-align` can be removed :) Your `.title` is a child to `.header` so it is aligned with flex aka `justify-content: center;`

ariallahyar (commented):
Nice work, Sofia! Your project is well structured and looks great.

Two suggestions for future work. 
1. Play around in Inspector to verify the code does something. It's really helpful to be able to ex. turn on and off different lines of css. It helps me a lot in figuring out things like flexbox and how divs relate to each other.
2. Commit often! Think of your work as “unsaved” as long as you haven’t committed. See my code for an example (screen clip of a few commits below). I commit every time I feel like I've achieved something. It's hard to remember at first but becomes good habit. :)
Screenshot 2022-02-02 at 20 37 36

PR #296, opened 2022-01-30
Arianna project-business-site

ariallahyar:
Here is a link to my deployed site:
https://supper-society.netlify.app/

I'm happy with how it looks on desktop and tablet, but need to work out some tweaks with the mobile view to remove accidental white space and try to adjust the "zoomed in" effect that happens after submitting the form.


code/style.css
```diff
@@ -1,7 +1,261 @@
...
+.radio-group {
+  display: flex;
+  justify-content: space-around;
+}
+
+.preference {
+  padding-right: 1rem;
+}
+
+.preference-option {
```
JaEngd, line 86: Text color inside the scroll down menu is set to blue as default in iOS.
Solve it by setting a color in the button tag (I think).

code/style.css
```diff
@@ -1,7 +1,261 @@
...
+}
+
+form button:hover {
+  color: rgb(0,51,51);
+  font-weight: bolder;
+  background-color: white;
+  cursor: pointer;
+}
+
+.submit-form {
```
JaEngd, line 115: Love this part! 

Would've been great to have a link to like a login form or something else that takes you back to the "main view".

code/style.css
ariallahyar: Thanks for pointing this out! I didn't notice it until after deploying, at which point I had no time to change it. I will look into it, and a tip to myself ns to deploy earlier to check iOS before it's too late!

code/style.css
ariallahyar: Good stretch goal suggestion, I hadn't thought of that. Thanks for the idea!

code/index.html
```diff
@@ -2,29 +2,92 @@
...
-    <!-- video or image as a header is cool :) -->
+    <!-- hero image header -->
+    <header>
+      <h1 class="header-title">Supper Society</h1>
+    </header>
+    <main>
+      <div class="container">
+        <!-- Signup form -->
+        <form id="form" action="https://httpbin.org/anything" method="post">
+          <h1 class="form-header">GET STARTED</h1>
```
JaEngd, line 21: For SEO purposes it's re

JaEngd (approved):
Hi Arianna,

Great work with your business site! I really enjoy the concept.

I can't find any bugs, it's responsive and works on different devices. You've met all the requirements on the Blue Level and some of the requirements on both red and black level.   

I really like the part when you submit your form and get a confirmation message. As well as the whole look which shows that you had a plan from the beginning and worked towards it. It seems that you have worked with mobile first also which is great.  

Since i'm not that good of reviewing code yet I have just made some small comments below. Let me know if you have any questions.

PR #298, opened 2022-01-30
Week 3 project - Nadia Lefebvre

nadialefebvre:


code/style.css
```diff
@@ -1,7 +1,452 @@
...
+    margin-bottom: 8rem;
+  }
+
+  .signup-wrapper {
+    margin-right: 1.5rem;
+  }
+
+  .accordion-wrapper {
+    margin-left: 1.5rem;
+  }
```
rijad90, line 446: Styling tip: make this the same hight as the signup

rijad90 (commented):
Just, wow Nadia! Really beautiful design and impressive work with the JS! 

"Unfortunately" I don't have much feedback for you beacuse you are way above my level! Keep it up 😃

PR #299, opened 2022-01-30
Week 3 Sara Boström Bundock

bostrombundock:


code/index.html
```diff
@@ -4,27 +4,54 @@
...
+     <label> First name </label>
+     <input id="" class="form" type="text" name="name" required>
+     
+     <label>Last name</label>
+     <input id="" class="form" type="text" name="last name" required>
+     
+     <label>E-mail</label>
+     <input class="form" type="email" name="E-mail"required>
+  
+    <button class="sendapplication" type="submit">Sign up for our baking course!</button>
```
EmmaSprings, line 30: Perhaps this class could be renamed "send-application" just to make it easier to read?

code/style.css
```diff
@@ -1,7 +1,145 @@
...
+font-family: Cambria, Cochin, Georgia, Times, 'Times New Roman', serif;
+background-color: cornsilk;
+
+}
+
+.header {
+height: 50vh;
+min-height: 50vh;
+width: 100vw;
+background-image: url("./images/macarons.jpg");
```
EmmaSprings, line 26: Nice! I didn't know you could add an image into the css! Learning new things everyday!

code/style.css
```diff
@@ -1,7 +1,145 @@
...
+ 
+}
+
+.accordion-container {
+  
+  display: flex;
+  flex-direction:column;
+  align-items: center;
+  text-align: center;
+  margin: auto;
```
EmmaSprings, line 95: maybe just use one line for margin and set it to 12px for the top in there, like: margin: 12px, 0 0 0; (Or what the auto would be) instead of having the margin-top: 12px; also. Just to decrease the number of lines with code :)

code/style.css
```diff
@@ -1,7 +1,145 @@
...
+.open +  p {
+  display: block ;
+  
+
+}
+  
+.footer {
+
+    text-align:right;
+    text-align:center
```
EmmaSprings, line 123: Missing a little semi colon ; after "text-align:center"

EmmaSprings (approved):
Well done Sara! This is a pretty website with the choices of colors and image! Good job with adding the "required" to the input fields and the color for valid vs invalid to the fields as well, really like that!

One thing I noticed when checking the responsiveness in the mobile version vs. The desktop version,  was that something is off with the width of the input and FAQ-containers.  It might be that the width is set to 30% in the main setting (for mobile), which works fine for the desktop version, but when I go to “iPhone X” for example, the container is too small for the content, so the input fields are going outside of the box. 
So instead of choosing width: 30%, I believe there are some other options that can be chosen. Maybe a set width like width: 300px; or so. However not sure this is the best option, I believe you can also increase the % of the width. Width: 80% looked good for example. Probably stating the obvious now, but this because the phone screen is much more narrow than desktop so 30% of 100% phone width is not much compared to desktop width. I think it could be a good way to set the size different depending on  the media query. 
I mainly tried to change things in the inspector tool to see the changes. 

I also believe that since the width is set to 30% in the main layout, it doesn’t change to 100% width in desktop mode, but I can be wrong here.

 Please ask if you have any questions about this comment! 

Overall you should definitely be proud of yourself and your work! Good job!!

code/style.css
bostrombundock: Aha, thanks, make completely sense. I had no idea about that!

bostrombundock:
Thank you Emma! So good and interesting inputs!

PR #304, opened 2022-02-03
Project business site week 3/Marianne Ardin

MarianneArdin:


code/script.js
```diff
@@ -1,9 +1,43 @@
...
+
+ document.getElementById('greeting').innerText= 'hello'
+}
+
+document.getElementById('form').onsubmit = event => {
+    event.preventDefault()
+  
+    // console.log('Form submittegit commit d')
+    // alert('Form submitted')
+    const inputValue = document.getElementById('nameInput').value
```
emmahogberg88, line 23: You have declared a const inputValue on both line 13 and 23, you only need to the declare the inputValue once, then you can use it again if you declared it as a global variable (e.i. not within  function) :)

code/style.css
```diff
@@ -1,7 +1,140 @@
...
+}
+
+/*This is the styling for the font*/
+body {
+  font-family:'Permanent Marker';
+  color: black;
+  background-color: aqua;
+}
+
+/*This is the styling for the header*/
```
emmahogberg88, line 19: Like that you are using pseudo code to divide your code in sections!

code/style.css
```diff
@@ -1,7 +1,140 @@
...
+h3 {
+  position: top;
+  padding: 1rem;
+  margin: 1rem;
+  color: black;
+  font-size: 20px;
+  font-style: underline;
+}
+
+form {
```
emmahogberg88, line 63: Maybe you can continue to add some /*comments in pseudo code*/ to divide your css in sections throughout the css file.

code/script.js
```diff
@@ -1,9 +1,43 @@
...
   this.classList.toggle("active")
 }
 
 // Selects an HTML element, and calls a function which will be executed when the element is clicked.
-document.getElementById("section1").onclick = toggle
+
+document.getElementById('form').onsubmit = (event) => {
+event.preventDefault()
+
+ const inputValue = document.getElementById('nameInput').value
```
emmahogberg88, line 13: Maybe you can gather all variables in the top of the document.

emmahogberg88 (commented):
Nice work Marianne! I like that your code is well structured and it´s easy to find what you are looking for. The comments you have added in some parts of the document is really nice and makes it easy to get a good overview, I suggest you continue with that 👍  

One suggestion is to add some more styling for the accordion to make it obvious for the
user that it´s possible to click at the headings (questions) and get an answer. Maybe by adding an underline, cursor: pointer; another background and/or a symbol that show that there is more to find :)    

When you are looking at your site on a big screen the accordion disappears a bit to the left. Maybe you could have put the form and accordion next to each other on big screen or have the accordion in center as you have with the form. 

Thanks for letting me review your code!

PR #312, opened 2022-08-17
Team Turtles group 1 - Kristin Cecilia Charlotte Amanda

Kristin-Larsson:
Task is to create a business one pager website that should contain:

✓ A header with responsive image/video
✓ A signup form with at least three different input types. Examples:
- Text fields
- A password field
- Set of radio buttons
- Set of checkboxes
- Submit button
✓ An FAQ accordion, built with JavaScript
✓ Style your page

Use a mob programming approach.


Sneezan:
**See it live here! :)** https://coruscating-sopapillas-e37efd.netlify.app

PR #366, opened 2024-08-30
Stars above

nella-x:
Website with responsive design and a form that has different input types.
I've tried to keep the code as clean and clear as possible, with good HTML structure.

https://stars-above.netlify.app/

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,4 +5,5 @@ I've created a business one-pager website that contains a hero header, a form an
 I used flexbox to place the content and make the design responsive. I made a form with different input types. For my future projects I want to keep exploring both flexbox and css grid. And I also want to try making different types of forms. When I made this site I coded the html first from top to bottom and then I styled it with CSS, starting with the smalest version (mobile first).
 
 ## View it live
-Every project should be deployed somewhere. Be sure to include the link to the deployed project so that the viewer can click around and see what it's all about.
+
+https://stars-above.netlify.app/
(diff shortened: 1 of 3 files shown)


HIPPIEKICK (approved):
Hi Nella, great job on creating this business site! Here comes some suggestions to think about going forward or tips to improve.

**HTML**
- It's great to see that you’re using a `<section>` tag for the workshops. This makes your content more structured and easier for search engines to index. Consider using a `<section>` tag for the sale-box as well.
- You’ve met the requirement of including multiple input types in your form. However, ensure that each input's attributes uses the correct quote characters. Comparison:`”` vs `"`, can you see that the first one is skewed and the second is more straight? Sometimes when we copy code, we get the skewed ones, but it's a complete different character, so make sure to always use the straight ones. To update your project, mark any `”` in Visual Studio Code and then do `cmd + D` (if you have Mac) to find all instances. That way you can easily replace them all. I think this will also "clean up" the response you get from HTTPBIN.

**CSS**
- Nice font you've found, and I think the design choices are well thought through.
- Ensure that your form is visually aligned with the rest of your website's design. Consider using padding, margins, and borders to make it stand out but remain cohesive with your overall design. A tip can be to try to align your elements to make it look nicer and make it more accessible.
- Removing the body's default margin will make more space for our content as well as making it look nicer.

Keep up the good work ⭐

nella-x:
Hello! Thank you for the feedback, I have made som adjustments. Nothing big. I'm not sure if the changes I've made to the form is what you requested, but I think it looks better now!

PR #375, opened 2024-09-01
project-2-business-site

erikamolsson:


HIPPIEKICK (dismissed):
Hi Erika, great job on your business site project! I would love to visit Scotland (by train though 🚂 😉).

**HTML**
- Your HTML is well-structured with clear sections for the header, main content, and form ⭐ 
- Your form is well-structured and clean as well, apart from the radio buttons. The value should be different for each input (as you've already written it), but the name needs to be the same for each input so that it's interpreted as a group. In your form now, it's possible to check all of the months, which is not the point with radio buttons.
- Keep the CSS to the CSS file, i.e. use CSS instead of the `<br>` tag

**CSS**
- The design looks clean! The consistent use of padding, margins, and a unified color scheme contributes to this. One thing I would suggest you to improve is the readability of the text on your hero image, as the contrast is not so strong.

**Changes requested**
- Group the radio buttons properly

Overall, you’ve done a really good job with this project. You've nailed the responsive design and included the necessary form elements. Just a few adjustments, such as grouping the radio buttons properly is needed. Keep up the good work!

JennieDalgren (approved):
⭐

PR #380, opened 2024-09-01
Mikas business site week 3

mikaeber:


HIPPIEKICK (approved):
Nice job with your business site, Mika! 🏃 You have a lot of files in your repo, so I'd suggest to remove all the ones you are not using. I started looking at demo.html before I realised it was a demo for the fonts 😅 Can be good to keep in mind going forward, to keep things clear and clean.

**HTML**
- The `br` tag is a bit outdated, try to achieve the same thing using flexbox
- I see a commented-out video tag for the hero section. It’s good practice to remove any unused code from your production files
- I see you've added "What's your main sport?" as one of the values in the select - great way of adding a placeholder! Just make sure to add the `disabled` attribute so that the user can't pick that option


**CSS**
- You’ve done a great job using CSS variables ⭐ Just make sure to remember to use them where applicable!
- Make sure your paths are correct. I think to reach the fonts from the CSS file you need the double dots to "move up a level" from the code folder - i.e. "../Fonts". And on this topic, rename all files and folders to lower case kebab-case.

Really good job overall with this project, just keep these things in mind as you progress 🚀

code/index.html
```diff
@@ -1,17 +1,84 @@
...
+          /><br /><br />
 
-    <!-- video or image as a header is cool :) -->
+          <input
+            type="email"
+            id="email"
+            name="email"
+            required
+            placeholder="email@example.com"
+          /><br /><br />
```
Jonash189: Using br for spacing in HTML can make the code harder to maintain and less flexible. It's better to handle layout and spacing with CSS, as it keeps the structure (HTML) and presentation (CSS) separated. This way, you can easily adjust spacing globally without having to modify the HTML itself. Consider using margin, padding, or flexbox/grid in your CSS to manage spacing between elements instead of relying on line breaks (br)

code/index.html
```diff
@@ -1,17 +1,84 @@
...
+              Reducing lactic acid by 12% and increases endurance by 8% on
+              average
+            </a>
+          </li>
+        </ul>
+      </nav>
+    </header>
+    <!-- <video autoplay muted loop id="hero-video">
+      <source src="/run.mp4" type="video/mp4" />
+    </video> -->
```
Jonash189: I think you forgot to change or delete this comment since you don’t have a video.

Fonts/demo.html
```diff
@@ -0,0 +1,192 @@
...
+            <p class="s36" style="font-size: 36px;">The quick brown fox jumps over the lazy dog.</p>
+            <p class="s48" style="font-size: 48px;">The quick brown fox jumps over the lazy dog.</p>
+            <p class="s60" style="font-size: 60px;">The quick brown fox jumps over the lazy dog.</p>
+            <p class="s72" style="font-size: 72px;">The quick brown fox jumps over the lazy dog.</p>
+        </div>
+    </div>
+
+</div>
+</body>
+</html>
```
Jonash189: What does this part of the code do? I'm curious.

Jonash189 (approved):
image

Great job! Your code works without any issues, and it looks really good on both mobile and desktop. I like the design, but one thing I would like to add is make (P) a bit more responsive on mobile devices with screen widths between 320 and 333 pixels. See image attacted image. 

 Overall, you have a nice structure for the HTML and CSS. It’s easy to follow and understand. Also noticed that you have a lot of code related to fonts, and I’m curious about what it does. Well done!

HIPPIEKICK (changes requested):
**Accessibility feedback**
Your form and the rest of the interactive elements are possible to navigate with keyboard, super! Your form have all the labels it needs too. You’ve used semantic elements, which means you get a lot of accessibility ”for free”, but you’ve also made use of aria labels to provide additional descriptive meaning to some of your elements.

I see you’ve worked on increasing the contrast for your colors this week, which is great - but the text "Get real with clinically tested isothiocyanates from 🥦"
is still really low contrast, so I’d suggest changing that to make it more accessible. Another thing to think about is the outline for form elements. The outline is there for a reason, so don’t style it away with too low of a contrast - it helps the user see what’s in focus 👀

**Changes Requested**
- Sufficient contrast for all text plus form elements' focus state

HIPPIEKICK (approved):

